PR #4, opened 2019-07-17
Rename to newspack-ads

jeffersonrabb:
The use case of this plugin has been expanded to cover ad services besides Google Ad Manager, so the plugin name has been genericized from `newspack-gam` to `newspack-ads`. This PR applies this change everywhere, in class names, function names, descriptions, etc. The block has been renamed to `Ad Unit` (slug: `ad-unit`). 

Previously when the block was named `newspack-gam/google-ad-manager` it was possible to add it to the block category "Newspack" because it had been established earlier in the load process by the plugin `newspack-blocks`. Now that this plugin's name comes first, it was necessary to replicate the block category code. This explains the changes to `webpack.config.js` and the `src/setup` directory. 

Companion PR for `newspack-plugin` is here: https://github.com/Automattic/newspack-plugin/compare/update/rename-newspack-ads?expand=1

#### Testing

1) Because the plugin has been renamed, it probably makes sense to delete the old repo and start clean.
2) In `wp-content/plugins/` of a test environment, clone this repo.
3) Checkout `update/rename-to-ads`
4) Run `npm install && npm run build:webpack`
5) `newspack-plugin` needs to be on `update/rename-newspack-ads` (and run `npm run build:webpack` there too)
6) All ad units created before these changes should exist as they were in the Newspack plugin UI.
7) Verify that they can be created, edited, and deleted just as before.
8) The block name has changed, so any previous instances of the block will be invalid and broken.
9) In a post, add new instances of `Ad Unit` block. Verify that these behave as they did previously.
10 Publish or preview the post and verify the ad behaves as before, with AMP and non-AMP codes.
11) Verify all tests pass.


newspack-ads.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Plugin Name:     Newspack Gds
```
philipjohn: Typo :)

newspack-ads.php
```diff
@@ -0,0 +1,22 @@
...
+ * Description:     Ad services integration.
+ * Author:          Automattic
+ * License:         GPL2
+ *
+ * @package         Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+// Define NEWSPACK_GAM_PLUGIN_FILE.
```
philipjohn: GAM remnant

src/blocks/ad-unit/index.js
```diff
@@ -24,12 +24,13 @@ export const icon = (
 		<Path d="M0 0h24v24H0z" fill="none"/>
 	</SVG>
 );
+console.log( getCategories() );
```
philipjohn: Leftover debug code

src/blocks/ad-unit/view.php
```diff
@@ -0,0 +1,51 @@
...
+		return '';
+	}
+
+	$classes = Newspack_Ads_Blocks::block_classes( 'wp-block-newspack-ads-blocks-ad-unit', $attributes );
+
+	$ad_unit = Newspack_Ads_Model::get_ad_unit( $active_ad );
+	$content = sprintf(
+		'<div class="%s">%s</div>',
+		esc_attr( $classes ),
+		$ad_unit['code'] /* TODO: escape with wp_kses() */
```
philipjohn: Just noting that kses will strip script tags

philipjohn:
Tested locally and everything is working as expected, with tests passing too.

philipjohn:
Oh I was going to do a separate issue for this, but I guess it is directly related to this PR. When you have an ad unit with small dimensions, it messes with the drop down in not-great ways. Here are two examples where I have an ad unit with 60px width or 60px height:

Screenshot from 2019-07-18 14-37-30
Screenshot from 2019-07-18 14-41-34

philipjohn:
Sorry, one more. Even with the https://github.com/Automattic/newspack-plugin/tree/update/rename-newspack-ads applied and built, I still see Google Ad Manager on the Ads screen:

Screenshot from 2019-07-18 14-46-33

newspack-ads.php
jeffersonrabb: Fixed https://github.com/Automattic/newspack-ads/pull/4/commits/eb62c583e263d2b68f7c50ab4a6bcd967acf2875

newspack-ads.php
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-ads/pull/4/commits/dbfbcbc47b76636d74d4dbb26a38dead7cc78641

src/blocks/ad-unit/index.js
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-ads/pull/4/commits/1d7cb0e105536b10046fe20fc928b277ee55507a

jeffersonrabb:
> Sorry, one more. Even with the https://github.com/Automattic/newspack-plugin/tree/update/rename-newspack-ads applied and built, I still see Google Ad Manager on the Ads screen:

This is expected and correct. These changes don't affect the UI.

jeffersonrabb:
> Oh I was going to do a separate issue for this, but I guess it is directly related to this PR. When you have an ad unit with small dimensions, it messes with the drop down in not-great ways. Here are two examples where I have an ad unit with 60px width or 60px height:

This isn't actually new in this PR—the move to having the dropdown in the block area was in the initial commit. I filed an issue for this here: https://github.com/Automattic/newspack-ads/issues/5

PR #6, opened 2019-07-18
Extend Ad Unit Data Model

jeffersonrabb:
This PR is an expansion of the data model for Ads. This includes:

- Support for AMP and non-AMP fields
- Support for different Ad providers (Google Ad Manager, etc)
- Header code stored as an option.

The header code will also be injected into the HEAD in non-AMP requests. This is meant to be tested along with https://github.com/Automattic/newspack-plugin/pull/190. Complete testing instructions can be found there.


jeffersonrabb:
Thanks for reviewing!

PR #10, opened 2019-09-13
Load Newspack Ads block scripts in widget blocks editor

claudiulodro:
Using the new Widget Blocks editor, Newspack Theme sidebars, and Super Cool Ad Inserter, we can insert arbitrary block-based content in the sidebar, footer, above-content, below-content, and in-content on all pages. This should meet our needs for all common scenarios: newsletter sign-ups, CTAs, ads, pop-up block, etc.

My main concern is that Widget Blocks is not super robust right now, but that might be something we can help with as the first people using it in production.

This PR adds some handling to load up the Ad Unit Block scripts in the Customizer experimental widget blocks screen, so that they will be available. We'll need to do something similar for any other custom blocks we want available in Widget Blocks.

To test:
1. Have the latest version of Gutenberg active and this PR.
2. Go to Customizer > Widget Blocks.
3. You should be able to add an Ad Unit block in the Widget Blocks (styling issues to be resolved later):
Screen Shot 2019-09-13 at 2 59 10 PM
4. Add an ad unit to a widget and view on frontend:
Screen Shot 2019-09-13 at 2 59 35 PM


claudiulodro:
This isn't necessary now that we have an ads widget.

Issue #14, opened 2020-01-08
Provide a widget, as well as a block


philipjohn:
While it's likely we can eventually just use blocks, at the moment widgets are needed to most effectively place ads.

To that end, we should add a widget with the same behaviour as the Ads block - providing a drop down box for the selection of a named Ad Unit as added in the Ads wizard.

Issue #13, opened 2020-01-08
Instead of requiring pasted code, take ad names and dimensions


philipjohn:
The plugin currently requires users to copy/paste ad code from Ad Manager. This can be a long process, and it's almost certainly a tedious process. When there is more than one ad (which is often) the global head code needs to be manually constructed from what GAM puts out for each individual ad - another long and tedious process.

Instead, it should be possible to ask for the following information only and construct ad code automagically:

* Google Ad Manager Network Code - this is the numerical ID for the GAM publisher account
* Ad Unit code - this is the machine-readable version of the "human" name for the ad unit, e.g. "banner_category_middle"
* Ad dimensions - This is the dimensions (e.g. 728x90) to show at each device breakpoint, enabling different size ads to be shown at different screen sizes

The plugin can then provide a block and a widget (#14), each of which provides a drop-down box for selecting the ad to insert based on the Ad Unit Name.

To output, the plugin needs to determine if it's outputting AMP-compatible HTML or not and accordingly construct the code. See pamTN9-kv-p2

# Google Ad Manager Network Code

This would replace the current Global Code setting in the Ads Wizard, and could look something like

![Screenshot from 2020-01-08 14-18-41](https://user-images.githubusercontent.com/136342/71985115-d78ef000-3221-11ea-89d4-98cc0d26c61d.png)

# Ad Unit code and Ad dimensions

This would replace the existing "Add/Edit Ad Unit" screen to ask for this information rather than HTML/AMP code. Something like this should work:

![Screenshot from 2020-01-08 14-01-43](https://user-images.githubusercontent.com/136342/71984118-22a80380-3220-11ea-9258-b788021c3a87.png)

philipjohn:
Closed via #15

PR #15, opened 2020-01-09
Ad Model Improvements

jeffersonrabb:
Complete testing instructions in https://github.com/Automattic/newspack-plugin/pull/375.

Closes https://github.com/Automattic/newspack-ads/issues/14, https://github.com/Automattic/newspack-ads/issues/13.

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -153,7 +153,7 @@ class Newspack_Ads_Blocks {
 		<script>
 			googletag.cmd.push(function() {
 				<?php foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) : ?>
-					googletag.defineSlot('/<?php echo esc_attr( $network_code ); ?>/<?php echo esc_attr( $ad_unit->post_title ); ?>', [ <?php echo esc_attr( implode( ',', $formatted_sizes[ $unique_id ] ) ); ?> ], 'div-gpt-ad-<?php echo esc_attr( $unique_id ); ?>-0').addService(googletag.pubads());
+					googletag.defineSlot('/<?php echo esc_attr( $network_code ); ?>/<?php echo esc_attr( $ad_unit->code ); ?>', [ <?php echo esc_attr( implode( ',', $formatted_sizes[ $unique_id ] ) ); ?> ], 'div-gpt-ad-<?php echo esc_attr( $unique_id ); ?>-0').addService(googletag.pubads());
 				<?php endforeach; ?>
 				googletag.pubads().enableSingleRequest();
 				googletag.enableServices();
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -140,11 +140,18 @@ class Newspack_Ads_Blocks {
 
 		$formatted_sizes = [];
 		foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) {
+			$sizes = $ad_unit->sizes;
+			usort(
+				$sizes,
+				function( $a, $b ) {
+					return $a[0] > $b[0] ? -1 : 1;
+				}
+			);
 			$formatted_sizes[ $unique_id ] = array_map(
 				function( $item ) {
 					return sprintf( '[%d,%d]', $item[0], $item[1] );
 				},
-				$ad_unit->sizes
+				$sizes
 			);
 		}
 
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -135,16 +135,31 @@ class Newspack_Ads_Blocks {
 		if ( $is_amp ) {
 			return;
 		}
+
 		$network_code = Newspack_Ads_Model::get_network_code( 'google_ad_manager' );
+
+		$formatted_sizes = [];
+		foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) {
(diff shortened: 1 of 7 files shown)


includes/class-newspack-ads-model.php
```diff
@@ -263,10 +264,10 @@ public static function sanitise_ad_unit( $ad_unit ) {
 		}
 
 		$sanitised_ad_unit = array(
-			'name'            => \esc_html( $ad_unit['name'] ),
-			self::AD_CODE     => $ad_unit[ self::AD_CODE ], // esc_js( $ad_unit['code'] ), @todo If a `script` tag goes here, esc_js is the wrong function to use.
-			self::AMP_AD_CODE => $ad_unit[ self::AMP_AD_CODE ], // esc_js( $ad_unit['code'] ), @todo If a `script` tag goes here, esc_js is the wrong function to use.
-			self::AD_SERVICE  => $ad_unit[ self::AD_SERVICE ],
+			'name'           => \esc_html( $ad_unit['name'] ),
+			self::CODE       => $ad_unit[ self::CODE ],
+			self::SIZES      => $ad_unit[ self::SIZES ],
```
claudiulodro: The code and sizes should be sanitized to some valid, reasonable value here (alphanumeric and `_` allowed for the code, and integer for the sizes). I wasn't able to hack a site in my testing (but it might be possible with the right string for someone more clever/better at JS than me). Currently inputting a width of e.g. `//` will cause havoc with the generated script tag in the header.

includes/class-newspack-ads-model.php
```diff
@@ -221,37 +226,33 @@ public static function delete_ad_unit( $id ) {
...
 	 * Retrieve the header code for a service.
 	 *
 	 * @param string $service The service.
-	 * @return string $header_code The code.
+	 * @return string $network_code The code.
 	 */
-	public static function get_header_code( $service ) {
-		return get_option( self::NEWSPACK_ADS_SERVICE_PREFIX . $service . self::NEWSPACK_ADS_HEADER_CODE_SUFFIX, '' );
+	public static function get_network_code( $service ) {
+		return get_option( self::NEWSPACK_ADS_SERVICE_PREFIX . $service . self::NEWSPACK_ADS_NETWORK_CODE_SUFFIX, '' );
```
claudiulodro: Sanitize this option before returning. People might extend this plugin in a child theme or something.

includes/class-newspack-ads-model.php
```diff
@@ -199,9 +207,6 @@ public static function update_ad_unit( $ad_unit ) {
 	 * @param integer $id The id of the ad unit to delete.
 	 */
 	public static function delete_ad_unit( $id ) {
-		if ( ! current_user_can( 'unfiltered_html' ) ) {
-			return false;
-		}
 		$ad_unit_post = \get_post( $id );
```
claudiulodro: Not a new issue from this PR, but while we're hardening the logic we should make sure the post type is ad unit so people can't delete arbitrary posts.

includes/class-newspack-ads-model.php
claudiulodro: For knowledge-sharing purposes, I'll expand on this. It's common for people to think that because the data is sanitized on the way in from this plugin, it will remain sanitized on the way out. This is not a safe assumption to make since all plugins and themes have access to the WP DB. The data could get modified by some other code after it's been saved, so we have to always assume everything straight from the DB is dangerous.

includes/class-newspack-ads-model.php
jeffersonrabb: This should do it: https://github.com/Automattic/newspack-ads/pull/15/commits/a9eaf915a22dc0e085161d7b9e3017dc431f5709

includes/class-newspack-ads-model.php
jeffersonrabb: https://github.com/Automattic/newspack-ads/pull/15/commits/d116e92900961962b98433db0ec5a9a3b2b1ea8e

includes/class-newspack-ads-model.php
jeffersonrabb: Size (and ad service type) sanitization in: https://github.com/Automattic/newspack-ads/pull/15/commits/4be47b8bb10476ce7ef116bfda0bae854a1829ce

Sanitizing the Code with `sanitize_title` in https://github.com/Automattic/newspack-ads/pull/15/commits/f6287fcbe7d9a8ea140bd198f3c293738ea45604

includes/class-newspack-ads-blocks.php
```diff
@@ -112,12 +113,58 @@ public static function enqueue_view_assets( $type ) {
...
 
-		if ( ! $is_amp ) {
-			// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
-			echo Newspack_Ads_Model::get_header_code( 'google_ad_manager' );
-			// phpcs:enable phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+		$formatted_sizes = [];
+		foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) {
+			$formatted_sizes[ $unique_id ] = array_map(
+				function( $item ) {
+					return sprintf( '[%s,%s]', $item[0], $item[1] );
```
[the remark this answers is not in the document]
jeffersonrabb: https://github.com/Automattic/newspack-ads/pull/15/commits/27b72e76cf03c3584ef0851b3680a537ced49cea

jeffersonrabb:
Good ideas! I addressed all with a variety of sanitization strategies.

PR #18, opened 2020-01-21
Update the block Placeholder style

thomasguillot:
This PR brings more consistency with the Newspack Plugin by using the same Material Icon as seen in NP > Advertising.

It also fixes an overflow issue with the Placeholder/Padding.

__Before:__

1 before

__After:__

Screenshot 2020-01-23 at 19 38 28


thomasguillot:
@jeffersonrabb I completely reworked this in 20de850 

Now the correct size/ratio is set to the placeholder and the select sits outside.

thomasguillot:
Closing this. I'll work on a different idea 💡

PR #19, opened 2020-01-23
Responsive Ads

jeffersonrabb:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

This PR approaches the the problem of responsive ads by rendering containers for each defined size and using media queries to control visibility. 

#### The Problem 

In Newspack plugin's Ads UI it is possible to define multiple sizes for a single unit. We incorrectly assumed that this would cause the appropriate creative to be served based on viewport size, but this functionality doesn't come for free. 

#### Research

[This doc](https://support.google.com/admanager/answer/7177589?hl=en) states:

> You can use CSS Media queries to define different ad sizes based on viewports

For non-AMP, [this doc](https://support.google.com/admanager/answer/3423562?hl=en&authuser=0) explains an approach to responsiveness using GPT `defineSizeMapping`. I wasn't enthusiastic about taking different approaches to AMP and non-AMP, so I opted to use CSS media queries in the non-AMP solution as well.

#### Approach

The idea here is that banner units should be treated as responsive (above header/below header/footer), while other ad units should not (widget, blocks, etc). This approach applies new rendering logic only to ad units in placements that suggest responsiveness. The rendering takes all of the sizes defined for the assigned ad unit and renders a separate container for each one. CSS is rendered programmatically that will hide/show each container based on viewport size. 

#### Open Questions

1. The documentation seems to fully support this approach for AMP, but I'm not 1000% confident that this is the right approach for non-AMP requests. The concern is that by defining multiple slots for the same unit the ad will be served more than once, and possibly counted.

2. If a user wished to use a different unit for mobile in the same location, we don't yet have an answer. 

3. The generated media queries assume the unit will be full width. In other words, if the defined sizes are 320x50 and 728x90, the resulting media queries would look like this:

```
@media (min-width:728px){}
@media (min-width:320px) and (max-width:728px){}
```
Is it appropriate that these match precisely to the size of the units, or should we account for any padding?

1. Create a Google Ad Manager ad unit with a typical desktop banner size (e.g. 728x90) and a typical mobile size (320x50). Use creative that will clearly show if the unit is appropriate for the viewport. Alternatively contact me or @philipjohn for sample codes to use.
2. Before switching any branches, ad this unit in Newspack plugin with both sizes defined. In Pre-defined ad placements, set the unit to display in `Global: Above Header`.
3. Verify that the ad is NOT responsive. Sometimes you'll see the mobile creative in desktop, and you should almost always see the desktop creative in mobile, overflowing the container.
4. Switch `newspack-ads` to this branch, and switch `newspack-plugin` to https://github.com/Automattic/newspack-plugin/pull/403.
5. After the switch the add should be responsive. You should see the correct creative for desktop and mobile. If you resize a window from one to the other you should see the creative change. 
6. Verify in both AMP and non-AMP requests.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you successfully ran tests with your changes locally?

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -140,7 +140,7 @@ class Newspack_Ads_Blocks {
 
 		$formatted_sizes = [];
 		foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) {
-			$sizes = $ad_unit->sizes;
+			$sizes = $ad_unit['sizes'];
 			usort(
 				$sizes,
 				function( $a, $b ) {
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
Thanks for reviewing on such short notice!

PR #31, opened 2020-03-19
feat: responsive ads for super cool ad inserter placements

jeffersonrabb:


jeffersonrabb:
In https://github.com/Automattic/newspack-ads/pull/31/commits/a35561347f27000fe4968041f8224b8284077baa I've simplified the logic so there is only one media query filter: `newspack_ads_media_query_for_size`

This is used to filter the media query for one size of an ad unit. The theme can alter the max/min width for a particular size based on theme considerations like sidebar presence and margins.

PR #40, opened 2020-04-30
Add support for ad targeting via filter

claudiulodro:
Part of https://github.com/Automattic/newspack-issues/issues/167

This PR adds support for adding targeting to ad units via a filter. Targeting is arbitrary key->val pairs when initializing an ad and can be used in GAM to fine-tune which creatives get delivered where.

In the future if other publishers want targeting, we can add UI for it to the ad management screens, but with little demand this should work fine.

To test:
1. Add the following code snippet somewhere:
```
add_filter( 'newspack_ads_ad_targeting', function( $targeting ) {
    $targeting['foo'] = 'bar';
    return $targeting;
} );
```
1 (alternate version). Get the custom plugin source [here](https://github.com/Automattic/newspack-issues/issues/167#issuecomment-622040752) and use the visual interface for adding targeting.
2. Examine the page source. In the non-AMP ads code you should see something like:
```
ad_units['banner_home_middle'] = googletag.defineSlot('/65246486/banner_home_middle', [ [ 300, 50 ] ], 'div-gpt-banner_home_middle-5eab192c2a166-300x50').addService(googletag.pubads());
ad_units['banner_home_middle'].setTargeting( 'foo', 'bar' );
```

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -172,13 +172,21 @@ class Newspack_Ads_Blocks {
 		?>
 		<script>
 			googletag.cmd.push(function() {
+				var ad_units = [];
 				<?php foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) : ?>
 					<?php if ( $ad_unit['responsive'] ) : ?>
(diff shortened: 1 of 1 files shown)


claudiulodro:
Thanks for the review!

PR #50, opened 2020-07-02
Don't define slots that aren't present, refactor non-AMP GAM script

claudiulodro:
The main purpose of this PR is to add a fix that will only define ad slots for ad containers that exist on the current page. This will prevent issues with and further optimize header bidding, ad delivery, etc. [That specific change is right here.](https://github.com/Automattic/newspack-ads/compare/fix/ad-slots?expand=1#diff-f00ee91a5e4c5891df399ba5eae186dcR198)

In order to make this change I had to do quite a bit of refactoring, because the special handling of responsive vs. regular ad units and the mixing of JS and PHP code made the codebase super difficult to work with. I've refactored the code in two ways:
1. Separate the PHP and JS from each other.
2. Within the JS section, simplify things to treat every ad slot the same way regardless of whether it's a responsive or regular slot.

It's difficult to follow along in the GitHub diff, but if you examine `Newspack_Ads_Blocks::insert_google_ad_manager_footer_code()` on its own, it should look much more straightforward than the previous implementation.

This is mostly a note for myself: BDN's Amazon UAM (header bidding) integration code will need to be slightly tweaked after this is merged, but the ad targeting integration and out-of-page-slot integration should continue working with no tweaks required.

### To test:
1. Set up the following ads: a global responsive ad, an ad block and/or widget.
2. Verify ads work nicely in non-AMP mode. There should be no visible changes to ad behavior.
3. (optional) You can view the page source to verify the information is correctly passed from PHP to JS. Check for the `ad_config` and `all_ad_units` variables in the page source.

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -151,59 +151,75 @@ class Newspack_Ads_Blocks {
 
 		$network_code = Newspack_Ads_Model::get_network_code( 'google_ad_manager' );
 
-		$formatted_sizes = [];
+		$prepared_unit_data = [];
 		foreach ( Newspack_Ads_Model::$ad_ids as $unique_id => $ad_unit ) {
-			$sizes = $ad_unit['sizes'];
-			usort(
-				$sizes,
-				function( $a, $b ) {
-					return $a[0] > $b[0] ? -1 : 1;
+			$ad_targeting = apply_filters( 'newspack_ads_ad_targeting', [], $ad_unit );
+
+			if ( $ad_unit['responsive'] ) {
+				foreach ( $ad_unit['sizes'] as $size ) {
+					$container_id = esc_attr( 'div-gpt-' . $ad_unit['code'] . '-' . $unique_id . '-' . absint( $size[0] ) . 'x' . absint( $size[1] ) );
+
+					$prepared_unit_data[ $container_id ] = [
+						'name'      => esc_attr( $ad_unit['name'] ),
+						'code'      => esc_attr( $ad_unit['code'] ),
+						'sizes'     => [ $size ],
+						'targeting' => $ad_targeting,
+					];
 				}
-			);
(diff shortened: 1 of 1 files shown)


claudiulodro:
> In this PR the code that sorted the sizes in descending order is removed: https://github.com/Automattic/newspack-ads/blob/master/includes/class-newspack-ads-blocks.php#L157-L162. Looking back, it seems this code is completely unnecessary since the sizes array for each unit ends up containing just one size. I think this was a relic of an earlier attempt at responsiveness that was abandoned because of limitations of how amp-ad works. Does this jibe with why you removed it?

The order of the sizes shouldn't matter at all when GAM determines which ads to display ([See step 4 here](https://support.google.com/admanager/answer/1143651)). [Responsive ads should be done like this example](https://developers.google.com/doubleclick-gpt/guides/ad-sizes#responsive_ads), and isn't something we are doing yet. (We will need to eventually, though)

> Should we add a toggle to the Newspack Ads Wizard to control disable_initial_load?

No, it really doesn't make sense for a user to control that. It's really just for further integrations, as disabling the initial load without doing anything else will just result in the ads not showing at all. Header bidding disables the initial load, and then displays ads once the bids have come in.

jeffersonrabb:
> The order of the sizes shouldn't matter at all when GAM determines which ads to display ([See step 4 here](https://support.google.com/admanager/answer/1143651)). [Responsive ads should be done like this example](https://developers.google.com/doubleclick-gpt/guides/ad-sizes#responsive_ads), and isn't something we are doing yet. (We will need to eventually, though)

Note that there isn't an AMP version of this logic, which is why we ended up doing it with media queries. We could have done separate implementations for AMP and non, but it seemed saner to use the same for both, even if it was the less elegant of the two approaches.

claudiulodro:
Added an issue about responsive ads so I don't forget about it :)  https://github.com/Automattic/newspack-ads/issues/51

PR #58, opened 2020-07-30
Automatic responsive ad handling for non-AMP mode.

### LINKED ISSUE #51: Responsive ads
We should follow the [GAM guide to implementing responsive ads](https://developers.google.com/doubleclick-gpt/guides/ad-sizes#responsive_ads). I think this could be done in a fully automated way using the existing Ads wizard with no new UI needed: For each breakpoint, add the configured ad sizes that are smaller than it. 

For example, if our mobile breakpoint is `780px`, and our ad sizes are `[[920,100],[320,100]]`, we would show both of those sizes on desktop and only the ones that fit on mobile e.g:

```
var mapping = googletag.sizeMapping()
  .addSize([780, 0],[[920,100],[320,100]])
  .addSize([0, 0], [[320,100]])
  .build();
```

There are a few other considerations, like not showing mobile ads on desktop, but this should at least prevent desktop ads from displaying on mobile. 

Related to #50, some further discussion there.

claudiulodro:
Closes #51 

This PR adds easy, no-config-necessary handling for responsive ads. It does this by making each unique ad size a responsive breakpoint, so that each ad size will only show on screens equal-to-or-wider than the ad size. This should work nicely for the vast majority of ad units with no further work on the publisher's part.

For example, an ad unit with sizes `[728, 90]` and `[320, 50]` will show ads of both sizes on screens >= `728px` and only `320x50` ads on screens smaller than `728px`.

### To test:
1. Configure an ad unit with multiple widths and add it to a post.
Screen Shot 2020-07-30 at 7 04 36 AM
2. Visit the post with desktop screen size and confirm the large ad gets displayed (the small ad will also get sometimes displayed).
Screen Shot 2020-07-30 at 7 06 43 AM
3. Shrink the screen size down. Refresh the page. Confirm only the small ad gets displayed.
Screen Shot 2020-07-30 at 7 07 15 AM

Note: Also added `supports` tag to prevent PHP notices on Gutenberg 8.6.0.

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -208,6 +208,34 @@ class Newspack_Ads_Blocks {
 					for ( var target_key in ad_unit['targeting'] ) {
 						defined_ad_units[ container_id ].setTargeting( target_key, ad_unit['targeting'][ target_key ] );
 					}
+
+					/** 
+					 * Configure responsive ads.
+					 * Ads wider than the viewport should not show.
+					 */
+
+					// Get all of the unique ad widths.
+					var unique_widths = {};
+					ad_unit['sizes'].forEach( function( size ) {
+						unique_widths[ size[0] ] = [];
+					} );
+
+					// For each width, get all of the sizes equal-to-or-smaller than it. 
+					for ( width in unique_widths ) {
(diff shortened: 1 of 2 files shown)


claudiulodro:
Thanks for the speedy review @jeffersonrabb!

Issue #55, opened 2020-07-20
Allow attributes to be added to Doubleclick AMP ad units' markup


dero:
Currently, the AMP markup for doubleclick ad units [has a firm structure](https://github.com/Automattic/newspack-ads/blob/e98b0fc0c573b80820802db50148f50af968f956/includes/class-newspack-ads-model.php#L390) that doesn't allow 3rd parties to change the generated markup (AFAIK, let me know if this is incorrect).

In our case we're running a native AMP site and would like to create a child theme and change the units' markup to potentially contain [`rtc-config`](https://github.com/ampproject/amphtml/blob/master/extensions/amp-a4a/rtc-documentation.md) and / or [`data-multi-size`](https://github.com/ampproject/amphtml/blob/master/extensions/amp-ad-network-doubleclick-impl/multi-size.md) attributes to increase ad demand competition, but we're unable to easily do so.

Would it be possible to [make the generated AMP ad unit string filterable](https://github.com/Automattic/newspack-ads/blob/e98b0fc0c573b80820802db50148f50af968f956/includes/class-newspack-ads-model.php#L397)? Or do you consider a filter like that to be too powerful and we should look for other options?

kmwilkerson:
Closing this issue as Newspack is in the process of moving away from AMP

PR #72, opened 2020-08-25
Optimize amp-ads for viewability.

adamsilverstein:
The `data-loading-strategy` attribute controls how eagerly (how many viewports) amp-ads load. See https://amp.dev/documentation/components/amp-ad/#data-loading-strategy-(optional)

This is important because if we load ads a user never scrolls to, these ads are never viewable and overall yield will suffer as a result. Unfortunately the default value is sub-optimal.

This article - https://medium.com/ampfuel/optimize-your-amp-pages-for-high-ad-viewability-rate-or-high-ads-served-311f6b539c73 - does a great job explaining the optimal value to use (and why). Fortunately the AMP runtime provides a constant we can use: `data-loading-strategy="prefer-viewability-over-views"`. This at the moment is 1.25 but may be fine tuned in the future if further testing reveals the need for a different number.

This PR ads the attribute by default with the expectation that it will improve yield for all users; let me know if you prefer this be controllable via a filter or if you plan on doing something broader like https://github.com/Automattic/newspack-ads/issues/55.

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -378,7 +378,7 @@ class Newspack_Ads_Model {
 		$largest = self::largest_ad_size( $sizes );
 
 		$code = sprintf(
-			'<amp-ad width=%s height=%s type="doubleclick" data-slot="/%s/%s"></amp-ad>',
+			'<amp-ad width=%s height=%s type="doubleclick" data-slot="/%s/%s" data-loading-strategy="prefer-viewability-over-views"></amp-ad>',
 			$largest[0],
 			$largest[1],
 			$network_code,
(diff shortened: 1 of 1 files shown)


claudiulodro (approved):
Sorry about the delay on reviewing this. Thanks for the contribution. This looks like a good improvement, and it works well in my testing. 

I think it makes sense to use this strategy by default instead of making things more complex for our users by requiring them to learn about loading strategies and make a decision on one. 👍

adekbadek:
Note: I've had to briefly turn off branch protection to merge this. The reason was that CI did not pick up this PR because "build from forks" setting was off. I've now turned it on, so this will not happen again.

PR #78, opened 2020-09-18
Add common ad targeting settings to ad units

claudiulodro:
This PR adds some common ad targeting to all ad units. This enables sites to fine-tune their ad delivery in Google Ad Manager:

- It adds category targeting for singles and category archives. This will let people target ads to only deliver on certain categories.
- It adds post/page slug and ID targeting for singles. This will let people target ads to only deliver on certain posts/pages or on the homepage.

These sort of ads are generally more lucrative for publishers than sitewide ads, because publishers can charge a premium on them, and at least a couple of our publishers have been selling ads in this way already.

### To test:

1. Configure and set a global ad unit.
2. With the network inspector open and **AMP enabled**, view a post in a category. Find the network request for the ad unit, scroll down to the `scp` query parameter, and observe category, slug, and ID ad targeting parameters were passed in the ad request:

Screen Shot 2020-09-18 at 9 57 30 AM

3. With the network inspector open and **AMP disabled**, view a post in a category. Find the network request for the ad unit, scroll down to the `prev_scp` query parameter, and observe category, slug, and ID ad targeting parameters were passed in the ad request:

Screen Shot 2020-09-18 at 9 57 49 AM

4. Configure and set an ad block.
5. Repeat steps 2 and 3 for the ad block:

Screen Shot 2020-09-18 at 9 59 19 AM
Screen Shot 2020-09-18 at 9 56 44 AM

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -539,10 +539,10 @@ class Newspack_Ads_Model {
 			$targeting['ID'] = get_the_ID();
 
 			// Add the category slugs to targeting on category archives.
-		} elseif ( 'WP_Term' === get_class( get_queried_object() ) ) {
-			$term_object = get_queried_object();
-			if ( 'category' === $term_object->taxonomy ) {
-				$targeting['category'] = [ sanitize_text_field( $term_object->slug ) ];
+		} elseif ( get_queried_object() ) {
+			$queried_object = get_queried_object();
+			if ( 'WP_Term' === get_class( $queried_object ) && 'category' === $queried_object->taxonomy ) {
+				$targeting['category'] = [ sanitize_text_field( $queried_object->slug ) ];
 			}
 		}
 
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -365,6 +365,7 @@ class Newspack_Ads_Model {
 		$sizes        = $ad_unit['sizes'];
 		$code         = $ad_unit['code'];
 		$network_code = self::get_network_code( 'google_ad_manager' );
+		$targeting    = self::get_ad_targeting( $ad_unit );
 		$unique_id    = uniqid();
 
 		if ( ! is_array( $sizes ) ) {
@@ -378,11 +379,12 @@ class Newspack_Ads_Model {
 		$largest = self::largest_ad_size( $sizes );
 
 		$code = sprintf(
-			'<amp-ad width=%s height=%s type="doubleclick" data-slot="/%s/%s" data-loading-strategy="prefer-viewability-over-views"></amp-ad>',
+			'<amp-ad width=%s height=%s type="doubleclick" data-slot="/%s/%s" data-loading-strategy="prefer-viewability-over-views" json=\'{"targeting":%s}\'></amp-ad>',
 			$largest[0],
 			$largest[1],
 			$network_code,
-			$code
+			$code,
+			wp_json_encode( $targeting )
 		);
 
 		return $code;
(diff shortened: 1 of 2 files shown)


includes/class-newspack-ads-model.php
```diff
@@ -508,5 +512,41 @@ function( $carry, $item ) {
...
+			$categories = wp_get_post_categories( get_the_ID(), [ 'fields' => 'slugs' ] );
+			if ( ! empty( $categories ) ) {
+				$targeting['category'] = array_map( 'sanitize_text_field', $categories );
+			}
+
+			// Add the post ID to targeting.
+			$targeting['ID'] = get_the_ID();
+
+			// Add the category slugs to targeting on category archives.
+		} elseif ( 'WP_Term' === get_class( get_queried_object() ) ) {
```
jeffersonrabb, line 542: I'm getting this PHP warning when editing a post:

```
PHP Warning:  get_class() expects parameter 1 to be object, null given in /srv/www/newspack-repos/newspack-ads/includes/class-newspack-ads-model.php on line 542
```

includes/class-newspack-ads-model.php
```diff
@@ -508,5 +512,41 @@ function( $carry, $item ) {
...
+			// Add the post slug to targeting.
+			$slug = get_post_field( 'post_name' );
+			if ( $slug ) {
+				$targeting['slug'] = sanitize_text_field( $slug );
+			}
+
+			// Add the category slugs to targeting.
+			$categories = wp_get_post_categories( get_the_ID(), [ 'fields' => 'slugs' ] );
+			if ( ! empty( $categories ) ) {
+				$targeting['category'] = array_map( 'sanitize_text_field', $categories );
```
jeffersonrabb, line 535: If the post is uncategorized, the param will be `category=uncategorized`. Is this as it should be? Or should we exclude the param in this case?

includes/class-newspack-ads-model.php
```diff
@@ -508,5 +512,41 @@ function( $carry, $item ) {
...
+		$targeting = [];
+
+		if ( is_singular() ) {
+			// Add the post slug to targeting.
+			$slug = get_post_field( 'post_name' );
+			if ( $slug ) {
+				$targeting['slug'] = sanitize_text_field( $slug );
+			}
+
+			// Add the category slugs to targeting.
```
jeffersonrabb, line 532: I tried this out on a post with the categories "Arts" and "Events & Concerts" - the result only includes the word "events":

```
slug=no-category&category=arts,events&ID=8076
```

includes/class-newspack-ads-model.php
jeffersonrabb: Update: I see the warning when loading the page on the frontend too.

jeffersonrabb (changes requested):
Left a few notes. Out of curiosity, why are the params different in AMP and non-AMP (`scp` vs. `prev_scp`)?

includes/class-newspack-ads-model.php
claudiulodro: Good catch. Should be fixed in https://github.com/Automattic/newspack-ads/pull/78/commits/dced583614bac06459f07a2f5683ba51981304b1

includes/class-newspack-ads-model.php
claudiulodro: This is correct. "Uncategorized" is still a category:

Screen Shot 2020-09-29 at 8 16 34 AM

A site could rename that category or pick a different default category (e.g. "News"), so we'd want it to be available for targeting.

includes/class-newspack-ads-model.php
claudiulodro: What's the category slug of your "Events & Concerts" category? If you simply renamed the Newspack default category "Events" to "Events & Concerts", it would keep the original slug of `events`, and this would therefore be the correct behavior.

claudiulodro:
Thanks for the review, @jeffersonrabb! I've made a revision and have one question.

includes/class-newspack-ads-model.php
jeffersonrabb: You're right - that was my slug. I created a new one with an ampersand and multiple words and the slug was assigned correctly - nothing to see here.

jeffersonrabb (approved):

PR #81, opened 2020-10-28
Try/amp plus gam

jeffersonrabb:
Full testing instructions in https://github.com/Automattic/newspack-plugin/pull/688

### DIFF
--- a/includes/class-newspack-ads.php
+++ b/includes/class-newspack-ads.php
@@ -72,5 +72,17 @@ final class Newspack_Ads {
 		}
 		return untrailingslashit( plugins_url( '/', NEWSPACK_ADS_PLUGIN_FILE ) ) . $path;
 	}
+
+	/**
+	 * Should current request be treated as an AMP endpoint.
+	 *
+	 * @return bool AMP or not
+	 */
+	public static function is_amp() {
+		if ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() ) {
+			return true;
+		}
+		return false;
+	}
 }
 Newspack_Ads::instance();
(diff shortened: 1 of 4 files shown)


includes/class-newspack-ads.php
```diff
@@ -72,5 +72,20 @@ public static function plugin_url( $path ) {
...
 		return untrailingslashit( plugins_url( '/', NEWSPACK_ADS_PLUGIN_FILE ) ) . $path;
 	}
+
+	/**
+	 * Should current request be treated as an AMP endpoint.
+	 *
+	 * @return bool AMP or not
+	 */
+	public static function is_amp() {
+		if ( apply_filters( 'should_use_amp_plus', 'gam' ) ) {
```
claudiulodro, line 82: See my feedback at https://github.com/Automattic/newspack-plugin/pull/688. It's confusing to expect a boolean response to `apply_filters( 'should_use_amp_plus', 'a string' )`. Traditionally you'd want something like `apply_filters( 'should_use_amp_plus', false, 'gam' )`.

As-is, unless something is hooked into this filter, it will return `'gam'` which means this check always evaluates to `true`.

claudiulodro (changes requested):
Works great! The targeting, responsive features, etc. all work nicely. Just one architectural thing I noticed that will prevent future headaches. :)

includes/class-newspack-ads.php
claudiulodro:
```suggestion
		if ( Newspack\AMP_Enhancements::should_use_amp_plus( 'gam' ) ) {
```

This is what this check would look like after my recommendation in https://github.com/Automattic/newspack-plugin/pull/688

jeffersonrabb: Done in https://github.com/Automattic/newspack-ads/pull/81/commits/84d90503bcf5e2747adffa4e564243f97f0008ee. I added a `method_exists` check as well to protect against fatals should `newspack-ads` update prior to `newspack-plugin`.

claudiulodro (approved):
Latest version looks good 👍

PR #82, opened 2020-10-29
Update block icon style and colour

thomasguillot:
Using Material's outlined instead of filled and add Newspack Blue colour to it.

The idea is to have our blue, just like Jetpack and WooCommerce do with respectively their green and purple.

Screenshot 2020-10-29 at 16 22 16


adekbadek (approved):
👍

PR #91, opened 2020-12-09
Improve responsive ad handling in non-AMP mode

claudiulodro:
This PR improves the way responsive ad handling works in non-AMP mode. It does so by using a somewhat simple heuristic:
- If an ad has sizes both smaller and larger than 500px wide, that ad is almost certainly a responsive ad with desktop and mobile sizes.
- On desktops, if there is a desktop-sized ad defined, don't show any mobile-sized ads.
- Otherwise, logic is the same as before.

This explanation makes more sense with some examples:
- Consider an ad with two sizes: `[[728, 90], [320, 50]]`. This is without a doubt, a responsive ad with a mobile and desktop size. On viewports >= 728px wide, we should display the `[728, 90]` ad. On viewports < 728px wide, we should display the `[320,50]` ad.
- Consider an ad with one size: `[[728,90]]`. This ad should only be shown on viewports >= 728px.
- Consider an ad with one size: `[[320, 50]]`. This ad should be shown on viewports >= 320px.

### To test
1. Set up an ad with `[[728, 90], [320, 50]]` sizes and set it as a global ad:

Screen Shot 2020-12-09 at 10 31 26 AM

2. On frontend, observe that the 728px ad displays on viewports >= 728px and the 320px ad displays on viewports < 728px. (**Note**: you need to refresh the page when changing the viewport size, since the ad is retrieved when the page first loads)

Desktop:

Screen Shot 2020-12-09 at 10 34 07 AM

Mobile:

Screen Shot 2020-12-09 at 11 07 47 AM

3. Delete the `[728, 90]` size. On frontend, observe the 320px ad displays on both mobile and desktop.

Screen Shot 2020-12-09 at 10 33 03 AM

Screen Shot 2020-12-09 at 10 32 55 AM

4. Delete the `[320, 50]` size and add the `[728, 90]` size again. On frontend, observe the 728px ad displays on desktop and no ad displays on mobile.

Screen Shot 2020-12-09 at 10 34 19 AM

Desktop:

Screen Shot 2020-12-09 at 10 34 07 AM

Mobile:

Screen Shot 2020-12-09 at 10 41 59 AM

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -217,9 +217,41 @@ class Newspack_Ads_Blocks {
 					// Build and set the responsive mapping.
 					// @see https://developers.google.com/doubleclick-gpt/guides/ad-sizes#responsive_ads
 					var mapping = googletag.sizeMapping();
-					for ( width in unique_widths ) {
-						mapping.addSize( [ parseInt( width ), 0 ], unique_widths[ width ] );
+
+					var mobile_cutoff = 500; // As a rule of thumb, let's say mobile ads are <500px wide and desktop ads are >=500px wide.
+
+					var smallest_ad_width = Math.min.apply( Math, Object.keys( unique_widths ).map( Number ) );
+					var largest_ad_width  = Math.max.apply( Math, Object.keys( unique_widths ).map( Number ) );
+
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Tested and works as described!

PR #95, opened 2021-01-07
Improved responsive AMP ad handling using multi-size

### LINKED ISSUE #87: Use multi-size ads for responsive AMP ads.
The responsive AMP ad handling renders an amp-ad element for each size and then hides/shows the appropriately-sized ad depending on the screen width. For a basic configuration with e.g. only two ad sizes 900x100 and 350x50, this works fine. For a more complex configuration with e.g. four ad sizes 900x100, 900x50, 350x100, 350x100, this approach does not work well. With such a configuration, the 900x50 and 350x50 ad sizes will never be displayed.

The markup for the above looks something like this:
```
<amp-ad width="900px" height="100px" ...></amp-ad>
<amp-ad width="900px" height="50px" ...></amp-ad>
<amp-ad width="350px" height="100px" ...></amp-ad>
<amp-ad width="350px" height="50px" ...></amp-ad>
```

A better approach for handling responsive AMP ads would be to add all sizes of the same width as [multi-sizes](https://github.com/ampproject/amphtml/blob/master/extensions/amp-ad-network-doubleclick-impl/multi-size.md) on each ad. The updated markup would look something like this:

```
<amp-ad width="900px" height="100px" data-multi-size="900x100,900x50"...></amp-ad>
<amp-ad width="350px" height="100px" data-multi-size="350x100,350x50" ...></amp-ad>
```

Additionally, some sites use ads with different widths that should probably display at the same breakpoint. It probably makes sense to aggregate these sizes together if the widths are close. For example, if a site has 300x250 and 320x250, either one of those should be a valid size on mobile viewports.

claudiulodro:
Closes #87 

This PR modifies the way we're generating responsive AMP ads to fix some issues when there are multiple ad sizes defined with the same width. See the example in #87 to better understand what I mean.

**Note**: I've removed the non-AMP parts out of `ad_elements_for_sizes`. Those bits were vestigial because they and aren't useful with the approach we've been using for non-AMP ads and that function never gets called in non-AMP mode.

## To test

(Ping me on Slack if you need specific GAM codes for these ads)

**1.** Use AMP Standard mode. Set up an ad with `300x250` and `300x50` sizes and assign it as a global header or footer ad:

Screen Shot 2021-01-07 at 11 00 03 AM

**2.** On the frontend, observe the network request that the ad makes. In the size param, there should be two sizes. There should only be one `amp-ad` element:

Screen Shot 2021-01-07 at 10 48 57 AM

**3.** Set up an ad with `400x50` and `300x250` sizes and assign it as a global header or footer ad:

Screen Shot 2021-01-07 at 11 00 20 AM

**4.** On the frontend, observe the network request that the ad makes. In the size param, there should be three sizes (one auto-generated based on the container size being `400x250`, and the two explicitly defined sizes). There should be two `amp-ad` elements (one for viewports >=400px with both sizes and one for viewports between 300px and 399px with the `300x250` ad):

Screen Shot 2021-01-07 at 11 02 30 AM

Screen Shot 2021-01-07 at 11 01 39 AM

**5.** Set up an ad with `970x90`, `728x90`, and `300x50` sizes and assign it as a global header or footer ad:

Screen Shot 2021-01-07 at 11 27 56 AM

**6.** On the frontend, there should be three `amp-ad` elements, none of which are multisize (The `970x90` for viewports >= 970px, `728x90` for viewports between 728px and 969px, and `300x50` for viewports between 300 and 727px)

Screen Shot 2021-01-07 at 11 34 07 AM

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -412,7 +412,7 @@ class Newspack_Ads_Model {
 		// As a heuristic, each ad slot can safely display ads 200px narrower or less than the slot's width.
 		// e.g. for the following setup: [[900,200], [750,200]], 
 		// We can display [[900,200], [750,200]] on viewports >= 900px and [[750,200]] on viewports < 900px.
-		$width_difference_max = 200;
+		$width_difference_max = apply_filters( 'newspack_ads_multisize_size_difference_max', 200, $ad_unit );
 		$all_ad_sizes         = [];
 		foreach ( $widths as $ad_width ) {
 			$valid_ad_sizes = [];
@@ -425,6 +425,7 @@ class Newspack_Ads_Model {
 
 			$all_ad_sizes[] = $valid_ad_sizes;
 		}
+		$all_ad_sizes = apply_filters( 'newspack_ads_multisize_ad_sizes', $all_ad_sizes, $ad_unit );
 
 		// Generate an array of media query data, with a likely min and max width for each size.
 		$media_queries = [];
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -390,6 +390,20 @@ class Newspack_Ads_Model {
 		return $code;
 	}
 
+	public static function prepare_multisizes( $sizes ) {
+		// Group all of the unique ad widths and their ad sizes.
+		$unique_widths = [];
+		foreach ( $sizes as $index => $size ) {
+			$width = absint( $size[0] );
+			if ( ! isset( $unique_widths[ $width ] ) ) {
+				$unique_widths[ $width ] = [];
+			}
+
+			$unique_widths[ $width ][] = $size;
+		}
+		return $unique_widths;
+	}
+
 	/**
 	 * Generate divs for a series of ad sizes.
 	 *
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
Looks good to me! Only thing I noticed is the third param is still present when `ad_elements_for_sizes` is called [here](https://github.com/Automattic/newspack-ads/blob/0b7da43b88e0e65ef4823a2206204b13b1efe9e4/includes/class-newspack-ads-model.php#L376), although it doesn't have any negative effect.

claudiulodro:
> Only thing I noticed is the third param is still present when ad_elements_for_sizes is called here, although it doesn't have any negative effect.

Good catch! I've fixed in https://github.com/Automattic/newspack-ads/pull/95/commits/fddcd0686a10de8bd2bdd0a21e5d7913c9cd5d8e

I also [added a couple filters for fine-tuning the sizes](https://github.com/Automattic/newspack-ads/pull/95/commits/a3b396f525b34daa31e7169deeb183c4993a9b20). This will come in handy for making minor, custom tweaks to the ad sizes for some sites (e.g. they want to show 728x90 and 970x90 ads at the same breakpoint, but that currently won't happen because the max size cutoff is 200px).

jeffersonrabb (approved):
Latest changes look good.

earcos:
Hi Claudiu, we're moving to Newspack very soon and will be using this plugin for ad inserts on AMP. But after doing tests on our banner ads and monetisation options, we found out that this approach is problematic for two reasons:

The plugin is are calling at least two ads using `<amp-ads>` but showing just one. This leads to:

1. **Unnecessary load**. Calling an ad that won't be shown. 
2. **Viewability issues**. You're calling a display ad that is not being shown on the page. Advertising is moving from a volume metric to a viewability metric. Since one of the banners is called but not shown, the viewability % of that ad is zero. That's a huge problem.

This could all be fixed using `data-multi-size` instead of inserting two or more `<amp-ads>`, one for each size. I believe there's a [PR request](https://github.com/Automattic/newspack-ads/pull/98) for  doing exactly this but it's been waiting to be reviewed for almost a month. Is there any way to make this change into production soon?

Thanks a lot!

claudiulodro:
Hi @earcos, thanks for the questions. Since amp-ads are lazy loaded, no ad call will be made for the ads which are hidden, and viewability should not be affected. For responsive amp-ads (where you have a small ad that is shown on mobile and a large ad that is shown on desktop), the only way to accomplish that with AMP is by having a mobile ad and a desktop ad and hiding/showing the appropriate ad based on the screen size. `data-multi-size` does not care about screen size, so just using that attribute with all the ad sizes is not enough (mobile-sized ads would display on desktop and vice-versa). 

That said, #98 does improve the size handling a lot so it should be merged and released very soon. :)

PR #98, opened 2021-01-19
Add multisizes to non-responsive AMP ads.

claudiulodro:
This PR addresses an issue where e.g. you set up an ad with sizes `300x250` and `300x600` and use that ad in a block or widget. Currently, only one of those ad sizes will ever be shown. Ideally, *either* of those two ad sizes would be shown. This PR implements this using the multi-size attribute, similar to #95, but much simpler because the ads don't have to be responsive. 

This should be fully backwards-compatible and shouldn't cause any issues on existing ad setups, except where someone may have set up an ad with multiple sizes and forgotten about it because only one size was ever shown. :) 

## To test:
1. Set up an ad with sizes 300x250 and 300x600, or you can use an existing responsive ad you have laying around (like the leaderboards from #95)

2. Assign the ad as a widget. On the frontend, inspect the element and network request and verify that all of the ad sizes are present as a `multi-size` attribute and in the network request (you should also get ads of all sizes if you're using real ad sizes):

Screen Shot 2021-01-18 at 4 15 48 PM

Screen Shot 2021-01-18 at 4 17 59 PM

3. Add a block with the ad unit. On the frontend, inspect the element and network request and verify that all of the ad sizes are present as a `multi-size` attribute and in the network request (you should also get ads of all sizes if you're using real ad sizes):

Screen Shot 2021-01-18 at 4 21 07 PM

4. (optional) If you want to verify backwards-compatibility, create an ad with only one size and assign it as a widget or add it as a block. There should be no multi-size, and it should continue working as before.

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -86,7 +86,7 @@ class Newspack_Ads_Model {
 
 			$prepared_ad_unit['ad_code']     = self::code_for_ad_unit( $prepared_ad_unit );
 			$prepared_ad_unit['amp_ad_code'] = self::amp_code_for_ad_unit( $prepared_ad_unit );
-			return $prepared_ad_unit;
+			return apply_filters( 'newspack_ads_ad_unit', $prepared_ad_unit );
 		} else {
 			return new WP_Error(
 				'newspack_no_adspot_found',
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works as described. When loading an ad unit with multiple sizes, the `sz` network request param shows all sizes. Also tested a single-size unit and saw the expected behavior.

Screen Shot 2021-02-19 at 1 09 16 PM

I will note that I only happened to see this one creative/size when testing a multisize unit—is that expected?

PR #101, opened 2021-01-27
fix: for sticky ads, use smallest ad unit size available

dkoo:
Right now, if an ad unit isn't responsive, we display only the largest available ad unit size. Sticky ads use `amp-sticky-ad`, which only supports a single child of `amp-ad`, and therefore can't be responsive. But the primary use case is for mobile devices, so we generally want to be showing the _smallest_ size available for that placement. This branch does exactly that, for sticky ads only. Other placements will still show the largest available size.

We may want to consider adding CSS to the theme to hide the sticky ad unit at viewports larger than mobile size—what do you think, @claudiulodro?

To test:

1. While on `master`, in the Newspack > Advertising wizard, enable Google Ad Manager and use the Newspack network code (ping me for the code if you need it).
2. Create an ad unit with the code `SecondBanner`. This unit is setup in our Google Ad account with two sizes: 375x46 and 780x95. Add both sizes in the wizard.
3. In Newspack > Advertising > Global Settings, enable the Sticky position and set it to use Google Ad Manager and the SecondBanner unit.
4. View the site front-end at a mobile viewport size and observe that the 780x95 creative is displayed in the sticky ad.
5. Check out this branch.
6. Refresh the page, now confirm that the 375x46 creative is displayed at all viewport sizes.

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -376,12 +376,12 @@ class Newspack_Ads_Model {
 			return self::ad_elements_for_sizes( $ad_unit, $unique_id );
 		}
 
-		$largest = self::largest_ad_size( $sizes );
+		$size_to_show = self::is_sticky( $ad_unit ) ? self::smallest_ad_size( $sizes ) : self::largest_ad_size( $sizes );
 
 		$code = sprintf(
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
🚢

PR #110, opened 2021-02-24
refactor: simplified approach for filtering out large sizes for sticky ads

dkoo:
I liked [your suggestion](https://github.com/Automattic/newspack-ads/pull/109#pullrequestreview-596930952) better than what I implemented, so I implemented and re-tested. Should show no difference in behavior compared to `master` now.


claudiulodro (approved):
:D

PR #113, opened 2021-02-26
fix: remove large sizes for sticky ads in non-amp

dkoo:
Remove large ad unit sizes for sticky ads, for non-AMP pages. Should basically do the same thing as #110 but for non-AMP.

Coupled with https://github.com/Automattic/newspack-theme/pull/1270 to hide the ad entirely at larger viewports. @claudiulodro, will this affect viewability of the ad unit? If so, how can we ensure that the sticky unit is never actually rendered at larger viewport sizes? Could use some help with that.

To test, follow instructions in https://github.com/Automattic/newspack-theme/pull/1270


adekbadek (commented):
LGTM, but awaiting @claudiulodro's answer to the question in PR description:

> will this affect viewability of the ad unit? If so, how can we ensure that the sticky unit is never actually rendered at larger viewport sizes?

I guess the sticky ad unit should be filtered out from the activated ads (depending on viewport size) in JS, somewhere [in here](https://github.com/Automattic/newspack-ads/blob/0167526f491ed4c8846650cbf4a501c181a6b63b/includes/class-newspack-ads-blocks.php#L189).

dkoo:
Adding @claudiulodro as a reviewer to weigh in on the viewability question.

dkoo:
@claudiulodro Any chance you could give this a look? It's the last remaining piece for sticky ads as they're currently implemented.

claudiulodro (approved):
Sorry on the delay! I didn't realize I'd been pinged so many times 😬 

Hiding an ad on non-AMP will affect the viewability, so we should avoid it. I've pushed up a small tweak to not initialize sticky ads on desktop in https://github.com/Automattic/newspack-ads/pull/113/commits/81469d019e62d088cde277971922dc79e5e9d659. This all works well in my testing. If it looks good to you this can be merged. :)

claudiulodro:
Bot automatically removed my Needs Review tag!

adekbadek (approved):

PR #119, opened 2021-04-05
Add PR labelling workflow

adekbadek:
So we have less clicking to do


dkoo (approved):
Let's test this right now ;)

PR #124, opened 2021-04-16
Add GAM integration

adekbadek:

The ad units created in GAM have to be manually recreated in Newspack, linking one with the other via the unique ad placement code. With changes here, the ad units are fetched from GAM and saved in the DB - in order not to query GAM on each page load.

The only downside is that the sync happens when the user edits the ad units, so the changes made in [GAM web UI](https://admanager.google.com/) will not be reflected until the user logs in to the Advertising wizard of Newspack plugin. This is not a regression though - it was like that before. 

1. On `master`, set up some ad units to show (via global placements or the Ad block) - to ensure the update will not break existing ad unit placements
1. Switch to this branch, and use `add/gam-ad-units-in-ads-plugin` branch of the Newspack plugin (https://github.com/Automattic/newspack-plugin/pull/936)
2. Do not authenticate via Google yet (or [revoke access](https://myaccount.google.com/permissions))
3. Visit the site, verify that the ads work as before
4. Visit the Advertising wizard - observe that legacy ad units are listed and editable
5. Set `NEWSPACK_ADS_USE_GAM` env var to value `true` - this enables the GAM connection
1. Authenticate Newspack to use Google data via Newspack dashboard (`NEWSPACK_GOOGLE_OAUTH_ENABLED` env var has to be set)
2. Observe that now the ad units like displays both legacy ad units, and GAM ad units
3. Create, edit, then archive a GAM ad unit - observe the changes are reflected in [GAM web UI](https://admanager.google.com/) at each step
4. Assign an ad unit to a global placement or insert via the Ad block - observe the ad unit displaying an ad* 
5. Observe the ad units set up while on `master` are still displayed
6. Change the network code in ~the options table (`_newspack_ads_service_google_ad_manager_network_code`)~ _edit: change value of `$user_network_code` [in `is_network_code_matched` method](https://github.com/Automattic/newspack-ads/blob/c30a9dd307f4aba23ad4e42d4591943e7189b8d3/includes/class-newspack-ads-model.php#L350)_ to something else. Observe a notice about the mismatch above ad units, and that the ad units editing and creation is unavailable in the UI

\* if properly configured - but in non-AMP context, an ad "surface" will be displayed anyway


includes/class-newspack-ads-gam.php
```diff
@@ -0,0 +1,386 @@
...
+				[
+					'AD_MANAGER' => [
+						'networkCode'     => '-', // Provide non-empty network code to pass validation.
+						'applicationName' => self::GAM_APP_NAME_FOR_LOGS,
+					],
+				]
+			);
+
+			$oauth2_credentials = self::get_google_oauth2_credentials();
+			$session            = ( new AdManagerSessionBuilder() )->from( $config )->withOAuth2Credential( $oauth2_credentials )->build();
```
dkoo, line 79: Does this have to be a separate session from the one used to interact with the API (created on L129)?

tests/unit-tests/class-newspack-ads-tests-model.php
```diff
@@ -1,112 +1,14 @@
...
+ * Tests the ads model functionality.
  *
  * @package Newspack\Tests
  */
 
 /**
- * Test plugin management functionality.
+ * Test ads model functionality.
  */
 class Newspack_Ads_Test_Plugin_Manager extends WP_UnitTestCase {
```
dkoo, line 11: I get why we no longer need most of these tests. Wonder if there are any tests we can add to cover the GAM interactions? e.g. checking the connection status, serializing the ad unit data (with mock data), syncing to the options table, etc.

dkoo (changes requested):
This is some great work and it's amazing to see it working with GAM already. I ran into a strange edge case that might be worth looking into, other than that everything worked great. Also had a couple of minor questions which are non-blocking.

Re: this:

> After a user who authorised with Google interacts with the Ad Units, there are synced from GAM to the local instance. There could be a situation, where another user, with access to a different GAM account, overwrites the Ad Units. This is because Google OAuth2 is per-user, while GAM Ad Units are per-site. A solution would be to make the Google connection per-site - this however introduces a security concern.

Might it be worth doing some functional testing around this case? e.g. a single test site with two users who are authenticated with two different GAM accounts. In terms of handling this scenario, wouldn't that manifest in a mismatch between the GAM network code the user has access to vs. the network code saved in the options table? So perhaps we can save synced data in the options table by network code so that we can at least avoid overwriting existing data, then only show users the data that matches their network code?

includes/class-newspack-ads-gam.php
adekbadek: Yes, this is a bit of the GAM PHP API weirdness - you need a network id in a session to get ad units (and everything else), but to get the network id you need another session 🤷‍♀️

includes/class-newspack-ads-gam.php
```diff
@@ -0,0 +1,386 @@
...
+			return false;
+		}
+	}
+
+	/**
+	 * Is GAM connected?
+	 *
+	 * @return boolean True if GAM is connected.
+	 */
+	public static function is_connected() {
```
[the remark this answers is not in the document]
adekbadek: > I think we need to also check whether the OAuth app is accessible instead of just checking for the credentials.

Good catch - added some changes in a4c7e43fb3d12b93d97e33014f3b0d875889d7b9. Now the backend will try to get the network code to verify the connection.

tests/unit-tests/class-newspack-ads-tests-model.php
adekbadek: Added tests in 63a58779b30435b9c877b78efbb7b6d4f02a69d6 👌

adekbadek:
> So perhaps we can separate the synced data in the options table by network code so that we can at least avoid overwriting existing data, then only show users the data that matches their network code?

Now the GAM properties are stored under the network code - in the future we might support switching networks. Also, if there's a network mismatch, user will see an error (see newly-added testing step 13.).

dkoo (approved):
Newest changes and tests look good!

adekbadek:
In order to selectively test these features, I've changed it so they are enabled by a feature flag. See updated testing instructions. Could you please review again @dkoo ?

dkoo (changes requested):
Great work! Ran into one minor PHP error, details below. Other than that, I experienced different (maybe better?) behavior at the final step:

> 12. Change the network code in the options table (`_newspack_ads_service_google_ad_manager_network_code`) to something else. Observe a notice about the mismatch above ad units, and that the ad units editing and creation is unavailable in the UI

I changed the option value to some garbage string, then refreshed the Advertising wizard. After refreshing, the option value seemed to revert back to the correct GAM network code and the front-end was unaffected (I could still see all ad units, create new ones, edit existing ones, etc.).

includes/class-newspack-ads-gam.php
```diff
@@ -0,0 +1,416 @@
...
+	 * If the AdUnit is not provided, create a new one.
+	 *
+	 * @param object $ad_unit_config Configuration for an Ad Unit.
+	 * @param AdUnit $ad_unit Ad Unit.
+	 * @return AdUnit Ad Unit.
+	 */
+	private static function modify_ad_unit( $ad_unit_config, $ad_unit = null ) {
+		$name   = $ad_unit_config['name'];
+		$sizes  = $ad_unit_config['sizes'];
+		$status = $ad_unit_config['status'];
```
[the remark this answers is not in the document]
adekbadek: Fixed in c30a9dd307f4aba23ad4e42d4591943e7189b8d3

adekbadek:
> Other than that, I experienced different (maybe better?) behavior at the final step:

@dkoo that's right, I forgot that the network code will be sychronised when the GAM data is loaded. To test network code mismatch, follow the updated step 12. in the PR description.

includes/class-newspack-ads-gam.php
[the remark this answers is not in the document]
dkoo: Confirmed, thanks!

dkoo (approved):
🙌

PR #129, opened 2021-05-26
Ensure only Ad Unit CPT are handled in query

adekbadek:
It turns out that another plugin can [modify](https://github.com/INN/link-roundups/blob/be0f2a93a892904bf1d55a9dfd8ccf9311822d87/inc/link-roundups/class-link-roundups.php#L60) the `WP_Query` ran by this plugin!

## How to test 

1. Install the [Link Roundups plugin](https://github.com/INN/link-roundups/)
2. Create at least one link roundup
3. On `master` visit Newspack->Advertising wizard and observe the link roundups listed as ad units
4. Switch to this branch, observe the problem gone 🪄

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -111,15 +111,16 @@ class Newspack_Ads_Model {
 
 		$query = new \WP_Query( $args );
 		if ( $query->have_posts() ) {
-			while ( $query->have_posts() ) {
-				$query->the_post();
-				$ad_units[] = array(
-					'id'             => \get_the_ID(),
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This does fix the issue, but can you explain why the query was picking up other post types despite the `post_type` arg on line 107?

adekbadek:
> but can you explain why the query was picking up other post types despite the post_type arg on line 107?

That's a mystery! I suppose `WP_Query` is using some globals.

adekbadek:
Same issue in Campaigns: https://github.com/Automattic/newspack-popups/pull/614

PR #146, opened 2021-07-15
Harden logic when no ad unit selected

claudiulodro:
This PR fixes an issue with the (soon-to-be-legacy) ads widget. If you save the widget without selecting any ad unit, it throws PHP warnings on the frontend. This is because in `Newspack_Ads_Model::get_ad_unit_for_display`, the call to `get_post( $id )` will return the current post when the default `$id` of `0` is passed in, and the current post has no ad meta data. I've hardened the logic to prevent this issue.

## To test/reproduce:
1. Before applying this patch, add an ads widget to the sidebar, don't select an ad unit, and Save:
Screen Shot 2021-07-15 at 11 12 56 AM
2. On the frontend, observe the warnings thrown in the sidebar:
Screen Shot 2021-07-15 at 11 09 51 AM
3. Apply this patch. Observe no warnings thrown in the sidebar:
Screen Shot 2021-07-15 at 11 14 25 AM

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -64,6 +64,16 @@ class Newspack_Ads_Model {
 	 * @return object Prepared ad unit, with markup for injecting on a page.
 	 */
 	public static function get_ad_unit_for_display( $id, $placement = null, $context = null ) {
+		if ( 0 === (int) $id ) {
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Nice catch! Can confirm that this fixes the issue.

PR #149, opened 2021-07-22
fix(editor): ad unit block

### LINKED ISSUE #5: Thin or Short Ad Units Lead to Ugly SelectControl
Noted by @philipjohn in https://github.com/Automattic/newspack-ads/pull/4#issuecomment-512822292. 
 
> When you have an ad unit with small dimensions, it messes with the drop down in not-great ways. Here are two examples where I have an ad unit with 60px width or 60px height:

adekbadek:
Closes #5 and fixes an issue that is probably WP 5.8-related.

I did think about moving the select to the editor sidebar, to accurately represent the super-small ad units in the block, but super-small ad units seem too much of an edge case to accommodate in such a way.

## How to test

1. On `master`,
1. Ensure you're on WP 5.8
2. Create 100x10 and a 10x100 ad units
3. Insert three Ad Unit blocks somewhere - choose the two units from 👆 and a regular one
4. Observe that the select fills the block on the regular ad unit
5. Observe that the block with the 10x100 ad unit is squished, and the select is a couple pixels thin only
6. Switch to this branch, rebuild, observe issues gone


miguelpeixe (changes requested):
The minimum width fixes the squished dropdown but creates an issue with the ratio logic for the 10x100 ad unit.

image


Although 10x100 is an unrealistic ad dimension, it might be a good idea to determine a limit for the ratio value. What do you think?

adekbadek:
Updated in 90c7ef9af2eb5461e68bb87cb0b79a81c7af90c0 - still, a 100x10 will show up as a square (as on your screenshot), because the minimum width of the placeholder in the block is 100px. But such sizes should not be common.

miguelpeixe (approved):
Nice! I like how this constraint looks:

image

PR #150, opened 2021-07-22
chore: remove grunt dependencies

adekbadek:
We're not using those.


miguelpeixe (approved):
`rm -r node_modules && npm ci && npm run build` runs successfully!

PR #151, opened 2021-07-22
fix: return all ad units

### LINKED ISSUE #106: Can't see all ad units if >100 ad units
We have a list of Ad units that we need to use in our website. At some point, it seems that there is a limit to the list of units. If I add a new ad unit, the plugin "erase" one of the list. We looked up and if we erase one of the list, a previous "deleted" ad unit came back. The impact that we have, is when we are placing the ad unit in the content, the dropdown list only shows the visible ads in the admin, so we are missing some units for specific pages.

It seems that the limit to show are 100 items. Any idea of how to show the complete list of ad units added?

adekbadek:
## How to test

Create >100 ad units, or trust [the docs](https://developer.wordpress.org/reference/classes/wp_query/#pagination-parameters) ;) :

> Use 'posts_per_page'=>-1 to show all posts (the 'offset' parameter is ignored with a -1 value).

Closes #106

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -141,7 +141,7 @@ class Newspack_Ads_Model {
 		$query           = new \WP_Query(
 			[
 				'post_type'      => self::$custom_post_type,
-				'posts_per_page' => 100,
+				'posts_per_page' => -1,
 				'post_status'    => [ 'publish' ],
 			]
 		);


miguelpeixe (approved):

claudiulodro:
Ah much simpler than I thought! :) I'd assumed it was using [the `per_page` param of the REST API, which does have a limit of 100 items.](https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/)

PR #156, opened 2021-08-17
feat: global add suppression settings

### LINKED ISSUE #155: Add control of ads on archive pages
HKFP got a notice from Google that some ads providers were opting not to display ads on their tag page because of the content type. They'd like to be able to turn off ads for that page to clear the notice in GAM but there is currently no easy way to do that.

Could we add a filter to our ad manager that allows for removing ads on specific archive pages in the same way that we do for individual posts?

adekbadek:
_Testing steps in https://github.com/Automattic/newspack-plugin/pull/1100_

Closes #155

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -12,8 +12,9 @@ class Newspack_Ads_Model {
 	const SIZES = 'sizes';
 	const CODE  = 'code';
 
-	const OPTION_NAME_NETWORK_CODE = '_newspack_ads_service_google_ad_manager_network_code';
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Works well, but see my [comment](https://github.com/Automattic/newspack-plugin/pull/1100#pullrequestreview-734505541) in the other PR.

adekbadek:
@dkoo - updated, see [comment](https://github.com/Automattic/newspack-plugin/pull/1100#issuecomment-902830456) in the other PR.

dkoo (approved):
Works great!

PR #168, opened 2021-09-02
Fix(targeting): remove post slug and add post type

### LINKED ISSUE #164: Remove slug key-val targeting
From @kmwilkerson at https://github.com/Automattic/newspack-ads/pull/78#issuecomment-907284430

> With the slug values, GAM limits it to 40 characters so those longer post titles will only be targetable within GAM using the first 40. For example, slug=programa-nc-hope-aun-acepta-solicitudes-de-ayuda-para-alquiler-con-servicio-en-espanol will have to be targeted as programa-nc-hope-aun-acepta-solicitudes- in GAM. Because of this, I think we should remove 'slug' as a targetable key so as not confuse publishers on why slug targeting doesn't work when they input the whole slug into the targeting picker. Using ID is the best way to go for single page/post targeting!

miguelpeixe:
As reported by @kmwilkerson, GAM limits targeting values to 40 characters so longer slugs will only be targetable within GAM using the first 40 characters. To avoid confusion for publishers, this PR removes `slug` as a targetable field and adds `post_type` for broader targetable coverage.

Closes #164.

### How to test

1. Make sure you have a global ad unit visible on a singular post page
2. With the network inspector open, visit a singular post and see make sure `category`, `post_type` and `ID` are being sent through the `scp` (with AMP enabled), or `prev_scp` (with AMP disabled) query parameter in the ad request.

Captura de Tela 2021-09-02 às 16 09 19

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -695,17 +695,14 @@ class Newspack_Ads_Model {
 		$targeting = [];
 
 		if ( is_singular() ) {
-			// Add the post slug to targeting.
-			$slug = get_post_field( 'post_name' );
-			if ( $slug ) {
-				$targeting['slug'] = sanitize_text_field( $slug );
-			}
-
 			// Add the category slugs to targeting.
 			$categories = wp_get_post_categories( get_the_ID(), [ 'fields' => 'slugs' ] );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described 👍

PR #163, opened 2021-08-27
feat(gam): connect using a service account

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

With the amount of red tape around a multi-site OAuth setup, it's easier to use Service Accounts.

After these changes, a GAM instance will be linked to Newspack not via OAuth flow, but via a Service Account flow:

1. Newspack creates a new service account
2. Publisher add the service account to their GAM instance 
3. Newspack places the JSON credentials file on publisher's WP instance

### How to test the changes in this Pull Request:

1. Create a Google Cloud Platform app and create a Service Account*
2. The service account has an email address. [Add it to a GAM instance](https://support.google.com/admanager/answer/6078734?hl=en).
3. Ensure Newspack's Google OAuth is turned off (or unset - `NEWSPACK_ADS_USE_GAM`)
4. Visit the Ads wizard, GAM configuration and observe the "legacy" view (the ad units should not have a toggle switch)
3. Download the JSON credentials of the service account (in "Keys" tab of service account settings) and place them on WP instance as `wp-content/google-service-account-creds.json` file
4. Reload the Ad wizard, observe the GAM ad units are there, with a toggle that de/activates them
5. Edit the ad units and observe the changes are synchronised to GAM 

\* IAM & Admin -> Service Accounts:

image

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #169, opened 2021-09-02
feat: allow multiple comma-delmiited GAM network codes

### LINKED ISSUE #157: Google MCM support
[Google MCM](https://support.google.com/admanager/topic/9470608) allows a "parent" GAM account to manage inventory for a "child" GAM account.

The ad code for this integration looks like this:

AMP:

```
<amp-ad width=970 height=250
    type="doubleclick"
    data-slot="/21759680430,1052128/70FM/MJL/MJL_leaderboard"
    data-multi-size="970x90,728x90,320x100,320x50,300x250">
</amp-ad>
```

non-AMP:

```
.defineSlot( '/21759680430,1052128/70FM/MJL/MJL_leaderboard', [[728, 90], [320, 100], [970, 250], [320, 50], [970, 90], [300, 250]], 'dfp-ad-leaderboard' )
```

Note in those codes that the format is `GAMACCOUNT1,GAMACCOUNT2[...]`, so it appears that our existing code will work if we can support a comma in the GAM network ID field:

Screen Shot 2021-08-19 at 1 00 26 PM

Currently sanitizing prevents the comma. We should allow numeric values and a comma.

dkoo:
Allows the advertising wizard to accept multiple GAM network codes as a comma-delimited string, as specified by MCM, and renders those codes in the slot attribute for both AMP and non-AMP ad units. Note that it's currently not possible to fully test MCM in a non-production environment.

Also fixes a minor bug in the Advertising wizard where saving a new value to the Network Code field and then navigating to a different view within the React app resets the field back to the previous value.

Closes #157.

### Multiple GAM network codes

1. Check out this branch as well as https://github.com/Automattic/newspack-plugin/pull/1123 (make sure to rebuild with `npm run build`).
2. In WP Admin, go to **Newspack > Advertising**. Turn on Google Ad Manager if not already enabled, then click "Configure."
3. In the **Network Codes** field, enter a single GAM code. **Note:** GAM codes must be integers, e.g. `12345`
4. Click "Save" and confirm that the wizard allows the single code to be saved.
5. In the **Network Codes** field, enter multiple GAM codes separated by commas and click "Save" again. Confirm that the wizard allows multiple comma-delimited codes to be saved.
6. Make sure you have at least one GAM ad unit defined here.
7. Go back to **Newspack > Advertising > Global Settings** and enable an ad unit in an ad position.
8. View the ad position on the front-end and confirm that the ad unit renders with multiple GAM network codes in both AMP and non-AMP contexts. Note that unless your test site and all GAM accounts are approved for Google's MCM program, the ad units will fail to render.
  * For AMP (test site must NOT have AMP Plus enabled): View page source, look for `<amp-ad>` elements, confirm that the `data-slot` attribute contains all comma-delimited network codes, e.g.

```html
<amp-ad width="1020px" height="230px" type="doubleclick" data-slot="/12345,67890/AdUnitName" data-loading-strategy="prefer-viewability-over-views" json='{"targeting":{"slug":"homepage","ID":1234}}' ></amp-ad>
```
  * For non-AMP: View page source, look for `var ad_config`, confirm that its `network_code` property contains all comma-delimited network codes.

### Saving and rendering Network Code field value

1. On `master`, go to **WP Admin > Newspack > Advertising** and configure Google Ad Manager.
2. Note the value in the Network Code field rendered on load.
3. Enter a new value, click save.
4. Click "Back to advertising options" then the "Global Settings" tab (basically mount a new React view without reloading the page). Then go back to **Ad Providers** and configure GAM again.
5. Observe that the value rendered in the Network Code field is the original value, not the newly saved value, until you refresh the page entirely.
6. Check out this branch and rebuild.
7. Repeat steps 1–4, but this time confirm that the value rendered when the view mounts matches the most recently saved value.

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -309,7 +309,7 @@ class Newspack_Ads_Model {
 	 */
 	public static function get_active_network_code() {
 		$network_code = get_option( self::OPTION_NAME_NETWORK_CODE, '' );
-		return absint( $network_code );
+		return sanitize_text_field( $network_code );
 	}
 
 	/**
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -359,7 +359,18 @@ class Newspack_Ads_Model {
 		}
 		try {
 			$user_network_code = Newspack_Ads_GAM::get_gam_network_code();
-			return absint( $user_network_code ) === $active_network_code;
+
+			// Active Network Code might be a comma-delimited list of codes.
+			return array_reduce(
+				explode( ',', $active_network_code ),
+				function( $valid_code, $code ) use ( $user_network_code ) {
+					if ( absint( $code ) === absint( $user_network_code ) ) {
+						$valid_code = true;
+					}
+					return $valid_code;
+				},
+				false
+			);
 		} catch ( \Exception $e ) {
 			// Can't get user's network code.
 			return false;


miguelpeixe (changes requested):
The implementation fails to handle a check for the authentication config file implemented by #163. It should also [change how the `is_network_code_matched` sanitizes the network code](https://github.com/Automattic/newspack-ads/blob/8e982cfe1bc0617bad2c4101f440100978448876/includes/class-newspack-ads-model.php#L362).

Besides that, everything is working as described! 🙌

dkoo:
@miguelpeixe Ah, I wasn't aware of the new GAM integration via Service Accounts. You're right that the implementation in this PR will only work if the site is using the "legacy" Ads experience, not connected to GAM in any way. But since this fix is a short-term requirement for a site that is currently not connected to GAM using a Service Account, I think we should update the UI to handle multiple connected GAM accounts in a separate PR. Issue opened here: https://github.com/Automattic/newspack-ads/issues/170

For now, I've updated the `is_network_code_matched` function to check whether the connected GAM code matches _one of_ the codes that's configured for the site. This means that in the interim, if a site is using the "legacy" Ads experience with multiple GAM codes and then subsequently adds Service Account credentials to connect to a GAM account, the synced GAM UI will work as long as the connected GAM account matches one of the saved codes.

miguelpeixe (approved):
Good idea! Thank you for revising, the method checks the network code correctly now.

PR #172, opened 2021-09-08
feat(gam): create target key-values on sync

### LINKED ISSUE #165: Create keys in GAM
When the user connects GAM, automatically create keys for the default set of Newspack keys used: ID and category. 

[API reference](https://developers.google.com/ad-manager/api/reference/v202108/CustomTargetingService?hl=en#createcustomtargetingkeys) for two required fields:
CustomTargetingKey.name for `ID` and `category`
CustomTargetingKey.type should be `FREEFORM`

miguelpeixe:
Automatically creates custom keys for [targeting key-values](https://support.google.com/admanager/answer/188092?hl=en). The created keys are `ID`, `category`, and `post_type` of type `FREEFORM` if they are not found `ACTIVE` in GAM.

Closes #165.

### How to test this PR

1. Visit your GAM dashboard at **Inventory -> Key-values** and make sure the keys mentioned above **don't exist**
2. Checkout this PR and make sure your setup is connected to GAM
3. Visit the Advertising wizard page (the custom keys should be checked and updated here, if necessary)
4. Visit your GAM dashboard and make sure the keys exists at **Inventory -> Key-values**

### DIFF
--- a/includes/class-newspack-ads-model.php
+++ b/includes/class-newspack-ads-model.php
@@ -171,8 +171,11 @@ class Newspack_Ads_Model {
 		if ( ! self::is_gam_connected() ) {
 			return $legacy_ad_units;
 		}
-		$ad_units = Newspack_Ads_GAM::get_serialised_gam_ad_units();
-		self::sync_gam_settings( $ad_units );
+		$ad_units    = Newspack_Ads_GAM::get_serialised_gam_ad_units();
+		$sync_result = self::sync_gam_settings( $ad_units );
+		if ( \is_wp_error( $sync_result ) ) {
+			return $sync_result;
+		}
 		return array_merge( $ad_units, $legacy_ad_units );
 	}
 
@@ -313,7 +316,7 @@ class Newspack_Ads_Model {
 	}
 
 	/**
-	 * Save GAM configuration locally and update GAM settings.
+	 * Save GAM configuration locally.
 	 * First reason is so the GAM API is not queried on frontend requests - information
 	 * about ad units & GAM settings will be read from the DB.
 	 * Second reason is backwards compatibility - in a previous version of the plugin,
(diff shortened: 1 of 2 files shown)


includes/class-newspack-ads-model.php
```diff
@@ -327,6 +327,8 @@ public static function sync_gam_settings( $serialised_ad_units = null, $settings
 			$serialised_ad_units = Newspack_Ads_GAM::get_serialised_gam_ad_units();
 		}
 		if ( null === $settings ) {
+			// Update targeting keys when fetching settings from GAM.
+			Newspack_Ads_GAM::update_custom_targeting_keys();
```
adekbadek, line 331: As with other interactions with the external service (GAM), this should all be in a `try/catch` block. Apart from that, it would be nice if the user was notified about this, as well as any errors, should they arise.

adekbadek (changes requested):

includes/class-newspack-ads-model.php
miguelpeixe: Good call! I've added `try/catch` for the sync method, which returns `WP_Error` that is handled by the `get_ad_units` method.

It required changes to the main plugin so it properly displays error notices and I've also moved the update execution to the `get_service()` method on the wizard class. This allows us to have better control over the state of the service on the UI.

To continue the review, also switch your `newspack-plugin` branch to this one: https://github.com/Automattic/newspack-plugin/pull/1146

You should remove the existing targeting keys so you can see the update in action.

In the case of targeting keys creation, the GAM card should display the following:
image

The link inside the notice goes to the Custom Targeting Keys managing dashboard on GAM.

adekbadek (approved):
👍

miguelpeixe:
I need another look to implement the changes related to https://github.com/Automattic/newspack-plugin/pull/1146:

 - eea61ee378420ac56e37f950a46edd65501a5396 implements the GAM setup method
 - b1668ff7ee6a1854b00014fb5cefdb9c55e5c698 changes the failed sync error data to not be interpreted as `notice` level by the wizard UI

adekbadek (approved):

PR #175, opened 2021-09-13
feat: add per-post suppression filter

### LINKED ISSUE #62: Use plugin functionality to hide ads on Sponsored Content
Right now we're working on a feature to allow posts to be marked as 'Sponsored Content' (either by being assigned a sponsored taxonomy, or by being assigned a specific sponsor). This option is being managed in a plugin that is currently under development. 

Part of this will include either fully hiding ads on sponsored content (or adding an option in the Sponsored Content plugin). 

Is it possible to either automatically toggle ads off on posts when they've been assigned a sponsored taxonomy, or a direct sponsor? 

image

adekbadek:
Closes #62

Follow steps in https://github.com/Automattic/newspack-sponsors/pull/62

### DIFF
--- a/includes/suppress-ads.php
+++ b/includes/suppress-ads.php
@@ -50,6 +50,8 @@ function newspack_ads_should_show_ads( $post_id = null ) {
 		if ( get_post_meta( $post_id, 'newspack_ads_suppress_ads', true ) ) {
 			$should_show = false;
 		}
+
+		$should_show = apply_filters( 'newspack_ads_should_display_for_post', true, $post_id );
 	}
 
 	$global_suppression_config = Newspack_Ads_Model::get_suppression_config();


miguelpeixe (changes requested):
The filter is sending and applying `true` although it can be false as per line 51.

I also think it might be a bit confusing to have 2 different filters being applied to the same variable. It can be hard to debug a `newspack_ads_should_display_for_post` being overwritten by `newspack_ads_should_show_ads`.

Perhaps we could return early for each query check logic or combine all query checks at the end of the function, relating the filter name to it? E.g.: `newspack_ads_should_{$current_query}`

adekbadek:
I totally missed that there's already a filter there, this PR is not needed 🤦. Updated https://github.com/Automattic/newspack-sponsors/pull/62

PR #178, opened 2021-09-16
feat(targeting): add slug back

adekbadek:
Partial revert of #168 - it has been reported that the `slug` is used by some.


kariae (approved):
LGTM

PR #184, opened 2021-09-21
Handle legacy network code and incompatible plugin

### LINKED ISSUE #183: Credential "network code" blocks editing
After uploading a GAM Service Account credential your UI gets stuck on "disabled editing" mode if you have an empty network code or a different code saved.

The UI should be updated to display the network code associated with the credential, using the same input but disabled. It should also display buttons to replace the credential and remove it, to return to legacy mode.

With this approach, Ad Units editing should never be disabled.

miguelpeixe:
Implements and closes #183

As suggested by the issue, the Ad Units editor will never be blocked, instead, the network code related to the credential "replaces" the saved network code. If the credentials are removed, the previous network code will be restored.

This change affects every method that uses `get_active_network_code()`, which includes ad units configuration methods.

This PR also introduces the `is_sdk_compatible()` method, which disables GAM connection and flags `incompatible` in the case Constant Contact Form plugin is enabled. 

### How to test this PR

1. Repeat the steps from https://github.com/Automattic/newspack-ads/pull/163 to create your Service Account and generate your credentials file
2. On a new installation, checkout https://github.com/Automattic/newspack-plugin/pull/1151, visit Newspack Ads dashboard and upload your credentials
3. Notice the network code set on the disabled input, with buttons to replace and remove credentials
4. Remove the credentials and save a random network code
5. Upload the credentials and observe the warning message

### Test incompatible plugin

1. Enable Constant Contact Form plugin
2. Visit Newspack Ads and observe the error notices and that the "upload credentials" button card is not visible

### DIFF
--- a/includes/class-newspack-ads-gam.php
+++ b/includes/class-newspack-ads-gam.php
@@ -496,11 +496,32 @@ class Newspack_Ads_GAM {
 	}
 
 	/**
+	 * Verify WP environment to make sure it's safe to use GAM.
+	 * 
+	 * @return bool Wether it's safe to use GAM.
+	 */
+	public static function is_sdk_compatible() {
+		// Constant Contact Form plugin loads an old version of Guzzle that breaks the SDK.
+		if ( class_exists( 'Constant_Contact' ) ) {
+			return false;
+		}
+		return true;
+	}
+
+	/**
 	 * Get GAM connection status.
 	 *
 	 * @return object Object with status information.
 	 */
 	public static function connection_status() {
+		if ( false === self::is_sdk_compatible() ) {
+			return [
+				'incompatible' => true,
+				'can_connect'  => false,
+				'connected'    => false,
+				'error'        => __( 'Cannot connect to Google Ad Manager.', 'newspack-ads' ),
+			];
+		}
 		$response = [ 'can_connect' => false !== self::get_service_account_credentials() ];
 		try {
 			$response['network_code'] = self::get_gam_network_code();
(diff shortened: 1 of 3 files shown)


dkoo (approved):
This handles the issue well and successfully avoids the "stuck" condition described in the issue, as well as the Guzzle conflict with Constant Contact Forms!

includes/class-newspack-ads-gam.php
```diff
@@ -495,16 +495,37 @@ function( $key ) {
 		return [];
 	}
 
+	/**
+	 * Verify WP environment to make sure it's safe to use GAM.
+	 * 
+	 * @return bool Wether it's safe to use GAM.
+	 */
+	public static function is_sdk_compatible() {
```
adekbadek, line 503: Nit: not sure about the name, I'd rather expect something like `is_environment_compatible`. In the future this might also check e.g. PHP version.

includes/class-newspack-ads-gam.php
```diff
@@ -495,16 +495,37 @@ function( $key ) {
...
 	 *
 	 * @return object Object with status information.
 	 */
 	public static function connection_status() {
+		if ( false === self::is_sdk_compatible() ) {
+			return [
+				'incompatible' => true,
+				'can_connect'  => false,
+				'connected'    => false,
+				'error'        => __( 'Cannot connect to Google Ad Manager.', 'newspack-ads' ),
```
adekbadek, line 522: It'd be nice to state the reason here 
```suggestion
				'error'        => __( 'Cannot connect to Google Ad Manager. This WordPress instance is not compatible with this feature.', 'newspack-ads' ),
```

adekbadek (approved):
Left suggestions

PR #202, opened 2021-10-18
perf: cache ad units

miguelpeixe:
GAM API has a rate limit of 8 requests per second, this PR implements a class variable to store fetched ad units to prevent excessive consecutive API calls.

## How to test changes

Make sure ad units are properly loaded and displayed on the Ads GAM wizard and Ad Unit widget configuration. Nothing should change.


dkoo:
Aha, that makes sense! Thanks for the additional info.

PR #213, opened 2021-10-29
feat: customizable settings

miguelpeixe:
Implements configurable settings for the Ads plugin. Initially set up to modify lazy load settings, but with support for multiple cases.

The default values for lazy load settings also changed and are as suggested by @kmwilkerson. It should produce the following immediate change on GPT configuration script:

```diff
googletag.pubads().enableLazyLoad( {
-  fetchMarginPercent: 500,   // Fetch slots within 5 viewports.
+  fetchMarginPercent: 100,   // Fetch slots within 1 viewport.
-  renderMarginPercent: 200,  // Render slots within 2 viewports.
+  renderMarginPercent: 0,  // Render slots upon view.
  mobileScaling: 2.0         // Double the above values on mobile.
} );
```

A setting is an array with the following keys:
- description: The description of the setting.
- help: The help text for the setting.
- section: The section the setting is in.
- key: The key of the setting. Should be used along with the section name.
- type: The type of the setting. Used to typecast the value.
- default: The default value of the setting.
- options: Options to be used for a select field. Values outside of this array will not update.
  - name: The name of the option.
  - value: The value of the option.
- public: Whether the setting value is allowed to be displayed publicly on the frontend.

Settings without `key` or with the `key` of value `active` should be interpreted as section headers on the UI. In the case of `active`, it is a module that can be activated or deactivated.

It also implements API endpoints for:
 - fetching settings list
 - fetching and updating settings values
 
## How to test

 1. Written tests should pass
 2. Test API at https://github.com/Automattic/newspack-plugin/pull/1208

### DIFF
--- a/includes/class-newspack-ads-settings.php
+++ b/includes/class-newspack-ads-settings.php
@@ -36,6 +36,8 @@ class Newspack_Ads_Settings {
 	 * - type: The type of the setting. Used to typecast the value.
 	 * - default: The default value of the setting.
 	 * - options: Options to be used for a select field. Values outside of this array will not update.
+	 *   - name: The name of the option.
+	 *   - value: The value of the option.
 	 * - public: Whether the setting value is allowed to be displayed publicly on the frontend.
 	 *
 	 * Settings without `key` or with the `key` of value `active` should be
--- a/includes/class-newspack-ads-settings.php
+++ b/includes/class-newspack-ads-settings.php
@@ -276,7 +276,7 @@ class Newspack_Ads_Settings {
 	 * @param string $section  The key for the section to update.
 	 * @param object $settings The new settings to update.
 	 *
-	 * @return object All settings.
+	 * @return object|WP_Error The settings list or error if a setting update fails.
 	 */
 	public static function update_section( $section, $settings ) {
 		foreach ( $settings as $key => $value ) {
--- a/includes/class-newspack-ads-settings.php
+++ b/includes/class-newspack-ads-settings.php
@@ -32,7 +32,7 @@ class Newspack_Ads_Settings {
 			self::API_NAMESPACE,
 			'/settings-list',
 			[
-				'methods'             => WP_REST_Server::READABLE,
+				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ __CLASS__, 'api_get_settings_list' ],
 				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
 			]
--- a/includes/class-newspack-ads-settings.php
+++ b/includes/class-newspack-ads-settings.php
@@ -12,17 +12,17 @@ defined( 'ABSPATH' ) || exit;
  */
 class Newspack_Ads_Settings {
 
-	const OPTION_KEY_PREFIX = '_newspack_ads_';
+	const OPTION_NAME_PREFIX = '_newspack_ads_';
 
 	/**
-	 * Get the setting key to be used on the options table.
+	 * Get the setting option name to be used on the options table.
 	 *
-	 * @param object $setting The setting to retrieve the key from.
+	 * @param object $setting The setting object to retrieve the key from.
 	 *
-	 * @return string Option key. 
+	 * @return string Option name. 
 	 */
-	private static function get_setting_option_key( $setting ) {
-		return self::OPTION_KEY_PREFIX . $setting['section'] . '_' . $setting['key'];
+	private static function get_setting_option_name( $setting ) {
+		return self::OPTION_NAME_PREFIX . $setting['section'] . '_' . $setting['key'];
 	}
 
 	/**
@@ -73,7 +73,7 @@ class Newspack_Ads_Settings {
 		$settings_list = array_map(
 			function ( $item ) {
 				$default       = ! empty( $item['default'] ) ? $item['default'] : false;
-				$item['value'] = get_option( self::get_setting_option_key( $item ), $default );
+				$item['value'] = get_option( self::get_setting_option_name( $item ), $default );
 				return $item;
 			},
 			$settings_list
@@ -103,7 +103,7 @@ class Newspack_Ads_Settings {
 		);
 		if ( $config ) {
 			settype( $value, $config['type'] );
-			return update_option( self::get_setting_option_key( $config ), $value );
+			return update_option( self::get_setting_option_name( $config ), $value );
 		} else {
 			return new WP_Error( 'newspack_ads_invalid_setting_update', __( 'Invalid setting.', 'newspack-ads' ) );
 		}
@@ -128,26 +128,36 @@ class Newspack_Ads_Settings {
 	}
 
 	/**
-	 * Get settings values organized by sections.
+	 * Get settings values grouped by sections.
 	 *
+	 * @param string  $section   The section to retrieve settings from.
 	 * @param boolean $is_public Whether to return only public settings.
 	 *
-	 * @return object Associative array containing settings values.
+	 * @return object Settings values or empty array if no values were found.
 	 */
-	public static function get_settings( $is_public = false ) {
+	public static function get_settings( $section = '', $is_public = false ) {
 		$list   = self::get_settings_list();
 		$values = [];
 		foreach ( $list as $setting ) {
 			if ( ! isset( $values[ $setting['section'] ] ) ) {
 				$values[ $setting['section'] ] = [];
 			}
-			if ( true === $is_public && true !== $setting['public'] ) {
+			// Skip non-public settings if specified.
+			if ( true === $is_public && ( ! isset( $setting['public'] ) || true !== $setting['public'] ) ) {
+				continue;
+			}
+			// Skip settings without key or value.
+			if ( ! isset( $setting['key'] ) || ! isset( $setting['value'] ) ) {
 				continue;
 			}
 			settype( $setting['value'], $setting['type'] );
 			$values[ $setting['section'] ][ $setting['key'] ] = $setting['value'];
 		}
-		return $values;
+		if ( $section ) {
+			return isset( $values[ $section ] ) ? $values[ $section ] : [];
+		} else {
+			return $values;
+		}
 	}
 
 }
--- a/includes/class-newspack-ads-settings.php
+++ b/includes/class-newspack-ads-settings.php
@@ -215,13 +215,20 @@ class Newspack_Ads_Settings {
 			),
 		);
 
+		$default_setting = array(
+			'section' => '',
+			'type'    => 'string',
+			'public'  => false,
+		);
+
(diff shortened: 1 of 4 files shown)


includes/class-newspack-ads-settings.php
```diff
@@ -0,0 +1,324 @@
...
+	 * - section: The section the setting is in.
+	 * - key: The key of the setting. Should be used along with the section name.
+	 * - type: The type of the setting. Used to typecast the value.
+	 * - default: The default value of the setting.
+	 * - options: Options to be used for a select field. Values outside of this array will not update.
+	 *   - name: The name of the option.
+	 *   - value: The value of the option.
+	 * - public: Whether the setting value is allowed to be displayed publicly on the frontend.
+	 *
+	 * Settings without `key` or with the `key` of value `active` should be
```
dkoo, line 172: I really like the idea of creating a more standardized way of structuring settings config for use by the Newspack wizard UI! I wonder if we could eventually port this to the main plugin and use it across all of our repos. Before that, maybe the API warrants some discussion?

For example, WP Core already has the concept of [settings sections](https://developer.wordpress.org/reference/functions/add_settings_section/) which are basically just configured with a slug and a display title—why not mimic this API instead of using a flat array structure and "special" or missing key values to denote the section? Might also make it a bit easier to group the settings by section up-front (as returned by the API endpoint), instead of handling it in JS using [`groupBy`](https://github.com/Automattic/newspack-plugin/pull/1208/files#diff-bde388cd74998cc3a84c1b3330379f7350c1dc99d7b73ee185833495fbdf39a5R66).

includes/class-newspack-ads-settings.php
```diff
@@ -0,0 +1,324 @@
...
+			self::API_NAMESPACE,
+			'/settings',
+			[
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ __CLASS__, 'api_update_section' ],
+				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+				'args'                => [
+					'section'  => [
+						'required'          => true,
+						'sanitize_callback' => 'sanitize_text_field',
```
dkoo, line 67: The values for these settings are numbers—should this be `absint`? Although if we want to use this same endpoint for any arbitrary settings that could be text or numbers, I suppose `sanitize_text_field` works.

dkoo (changes requested):
I love this feature—I can really see a use for the more standardized/abstracted settings config you implemented here. See below for a discussion question I had about that.

I also confirmed that the lazy load settings are outputted to the `googletag.pubads().enableLazyLoad` config object, but for some reason I can't see any ad units at the moment, so I'm not able to fully confirm the lazy-loading behavior. 😬

includes/class-newspack-ads-settings.php
miguelpeixe: In this specific case, it receives the section name, which is always a string. The settings data are sent through the `settings` arg and is sanitized [here](https://github.com/Automattic/newspack-ads/pull/213/files#diff-52802592e1cb377b15ce60d7ab90c75f6ae26f66be53709ea3c03b04aca7835fR104-R113), using `$config['type']` for typecasting, which I take suggestions on improvements!

miguelpeixe: `absint` does not work for negative or float values. For lazy load specifically it is only an issue for the `mobile_scaling` setting, which is `float`, but in abstraction terms, it is not ideal for our settings.

includes/class-newspack-ads-settings.php
miguelpeixe: I thought about this too! After testing different scenarios and different plugins, it can be part of a plugin module registration logic.

I'm also taking suggestions on its model. My thoughts on the section system are towards simplification, we are currently not *registering* these, like the WP Core implementation. [As you can see how the UI currently  implements it](https://github.com/Automattic/newspack-plugin/blob/eb39b664dab93af99dcb7114cc6854d0f664361f/assets/wizards/advertising/views/settings/settings-section/index.js), this solution handles well when a section is a "module" that can be activated. If we decide that the `groupBy` after fetch is not ideal, the grouping can live on the backend. Perhaps a different API endpoint or an endpoint argument. WDYT?

miguelpeixe:
81f6fa829f9d5d35e89203a33862d11d3172b578 implements sanitization specific for storing values on the `wp_options` table.

[As experienced here](https://github.com/Automattic/newspack-plugin/pull/1208#discussion_r743228143), cases in which the boolean setting defaults to `true` and will be stored for the first time with the value `false` clashes with a [known issue](https://core.trac.wordpress.org/ticket/40007) with the `update_option()` function.

To overcome this issue, boolean values will be casted into integer.

includes/class-newspack-ads-settings.php
miguelpeixe: d9c8e6518decf1f95d0dd17c333fc7726d1499b6 along with https://github.com/Automattic/newspack-plugin/pull/1208/commits/4ab2f2f3036f69d9461febd577b72617d59ea741 implements associative array support on the `get_settings_list()` method and use as default for the API endpoint

dkoo: I get you. One idea for shadowing the core settings section API is that it might allow us to "detach" these settings from the main plugin/wizard UI, so that if we don't have the wizard, we could always display a more Core-friendly settings page using the same config arrays. But since that might not be the direction we go, I'm okay keeping the current model for the simplicity.

includes/class-newspack-ads-settings.php
```diff
@@ -215,18 +221,40 @@ public static function get_settings_list() {
...
 		$settings_list = apply_filters( 'newspack_ads_settings_list', $settings_list );
 
+		// Add default settings and get values.
 		$settings_list = array_map(
-			function ( $item ) {
-				$default       = isset( $item['default'] ) ? $item['default'] : false;
-				$item['value'] = get_option( self::get_setting_option_name( $item ), $default );
-				settype( $item['value'], $item['type'] );
+			function ( $item ) use ( $default_setting ) {
+				$item          = array_merge( $default_setting, $item );
```
dkoo, line 235: Non-blocking: `wp_parse_args`? It won't do anything different in this case, but I tend to stick with that for merging options arrays with defaults.

dkoo (approved):
Thanks for the extra sanitization and the option to get an associative or indexed array of settings—I think that might be useful in the future.

includes/class-newspack-ads-settings.php
miguelpeixe: Thank you for your considerations! I'm happy to revisit this once we start implementing it in other plugins. Even though it is a structural part of the concept, it doesn't affect how data is stored and shouldn't require too much rework and no backward compatibility to adapt to a different approach.

How I'm seeing it now is that we can get the configuration from `get_settings_list()` and build the core friendly `add_settings_section()` and `add_settings_field()` in the case of a standalone setup while we work our way into modularizing the wizard js so it can live in the plugin's repository.

Although it could work the other way around, I like the idea of setting the parameters ourselves so it can better communicate with how the wizard and its components are put together.

includes/class-newspack-ads-settings.php
miguelpeixe: Good call! Done in a2307b1c6ce35cfa3a023522370a6f606217666c

dkoo:
> [As experienced here](https://github.com/Automattic/newspack-plugin/pull/1208#discussion_r743228143), cases in which the boolean setting defaults to `true` and will be stored for the first time with the value `false` clashes with a [known issue](https://core.trac.wordpress.org/ticket/40007) with the `update_option()` function.

TIL.

dkoo (approved):
Looks good! 🎉

includes/class-newspack-ads-settings.php
dkoo: Sounds good to me—I do appreciate the value of structuring it for the primary use case (the wizard UI) rather than the fallback (core WP settings). Thanks for talking it through!

includes/class-newspack-ads-settings.php
```diff
@@ -0,0 +1,374 @@
...
+	 * Update setting section.
+	 *
+	 * @return WP_REST_Response containing the settings list.
+	 */
+	public static function api_get_settings_list() {
+		return \rest_ensure_response( self::get_settings_list( true ) );
+	}
+
+	/**
+	 * Update setting section.
```
adekbadek, line 125: Description should be updated.

includes/class-newspack-ads-settings.php
```diff
@@ -0,0 +1,374 @@
...
+		foreach ( $settings as $key => $value ) {
+			$config = self::get_setting_config( $section, $key );
+			settype( $value, $config['type'] );
+			$sanitized_settings[ $key ] = $value;
+		}
+		return $sanitized_settings;
+	}
+
+	/**
+	 * Update setting section.
```
adekbadek, line 116: Description should be updated.

includes/class-newspack-ads-settings.php
```diff
@@ -0,0 +1,374 @@
...
+				'callback'            => [ __CLASS__, 'api_get_settings_list' ],
+				'permission_callback' => [ __CLASS__, 'api_permissions_check' ],
+			]
+		);
+
+		register_rest_route(
+			self::API_NAMESPACE,
+			'/settings',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
```
adekbadek, line 45: I find it confusing that there are two ways of `GET`ting the settings list – one with section data (`/settings-list`), other "raw" (`/settings`). Also only the former is used in the proposed settings consumer (https://github.com/Automattic/newspack-plugin/pull/1208), which leaves the latter without a use case. 
I'd suggest removing `/settings` route and the associated code, and renaming `settings-list` to `settings`. This would make for a clearer API with a `/settings` endpoint that handles `GET`s and `POST`s.

adekbadek (commented):
A bit late to the party, I know.

includes/class-newspack-ads-settings.php
miguelpeixe: See https://github.com/Automattic/newspack-ads/pull/220

includes/class-newspack-ads-settings.php
miguelpeixe: See https://github.com/Automattic/newspack-ads/pull/220

includes/class-newspack-ads-settings.php
miguelpeixe: Good call! My thoughts were to reproduce the methods logic, which is for the `get_settings_list()` -> `/settings-list` to return an array with metadata for UI and the regular `get_settings()` -> `/settings` to return raw data object.

Since we don't have a use case for the initially proposed `/settings` endpoint, I agree we should simplify this and later revisit if its the case for a different approach.

Done in https://github.com/Automattic/newspack-ads/pull/220

PR #214, opened 2021-11-01
Fix/circleci build

miguelpeixe:
Recent changes made to `circleci/php:7.3-node-browsers` resulted in our builds fail due to an incompatibility with our `node-sass` dependency.

[`circleci/` images are being deprecated and will lose support starting December 2021](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034). We should start moving towards maintained images.

This PR implements the use of `cimg/` images and changes the PHP extension install method from the `docker-php-ext-install` command to `apt` since the command is not supported on `cimg/php:7.4`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ commands:
       - run:
           name: Install PHP extensions
           command: |
-            sudo apt-get update && sudo apt-get install -y libxml2-dev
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Builds seem fine, would be nice to test an alpha build as well for the release tasks.

Screen Shot 2021-11-01 at 1 53 09 PM

PR #216, opened 2021-11-02
feat(gam): service account credentials managing

adekbadek:
Reinstates Service Account connection management (removed in ad5a5ac4dc6c59602f9b71f4ea78c337e3c55507), if OAuth is unavailable.

If both OAuth and Service Account connections are set up, the former will be preferred. 

See https://github.com/Automattic/newspack-plugin/discussions/1190


miguelpeixe (changes requested):
Working as described!

I don't think we should put the Service Account method behind an environment variable. It is the proposed default behavior for the plugin when OAuth is not available.

It should rather verify if OAuth is configured to not display or implement the method.

adekbadek:
> It is the proposed default behavior for the plugin when OAuth is not available.

Brilliant! Somehow I've missed it. Done in 569ef7b23676b7e24d4d66ffa58101e675b5d112

miguelpeixe (approved):
Working nicely!

PR #224, opened 2021-11-12
feat: integrate SCAIP settings

### LINKED ISSUE #177: Integrate SCAIP settings into ad wizard
The goal is to integrate the functionality/plugin settings of the SCAIP into our ad wizard. 

1. The fields that currently live under Settings > Ad Inserter should be moved to Newspack > Advertising > Global Settings
2. Selection of what ad units should be injected in each placement currently lives in the widget settings making it a very disjointed experience. This should also be moved to Global Settings, as it would seamlessly fit the flow of options there.

miguelpeixe:
image

This PR adds SCAIP settings as a section on Ads settings.

I'd appreciate a second opinion from the reviewer if we should also remove the original SCAIP settings page. This is not implemented here and I'm not sure if we should.

Closes #177.

### How to test

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1248 for a section layout fix
2. Make sure you have SCAIP installed and have tweaked its default settings
3. Visit the Advertising -> Global Settings wizard
4. Observe your settings are being shown at the "Post ad inserter settings"
5. Update the settings, visit the SCAIP settings page and confirm the results match

### DIFF
--- a/includes/class-newspack-ads-scaip.php
+++ b/includes/class-newspack-ads-scaip.php
@@ -1,0 +1,97 @@
+<?php
+/**
+ * Newspack Ads SCAIP Hooks
+ *
+ * @package Newspack
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Newspack Ads SCAIP Class.
+ */
+class Newspack_Ads_SCAIP {
+
+	// Map of SCAIP option names.
+	const OPTIONS_MAP = array(
+		'start'          => 'scaip_settings_start',
+		'period'         => 'scaip_settings_period',
+		'repetitions'    => 'scaip_settings_repetitions',
+		'min_paragraphs' => 'scaip_settings_min_paragraphs',
+	);
+
+	/**
+	 * Initialize SCAIP Hooks.
+	 */
+	public static function init() {
+		add_filter( 'newspack_ads_settings_list', array( __CLASS__, 'add_settings' ) );
+		add_filter( 'newspack_ads_setting_option_name', array( __CLASS__, 'map_option_name' ), 10, 2 );
+	}
+
+	/**
+	 * Add SCAIP settings to the list of settings.
+	 *
+	 * @param array $settings_list List of settings.
+	 *
+	 * @return array Updated list of settings.
+	 */
+	public static function add_settings( $settings_list ) {
+
+		if ( ! defined( 'SCAIP_PLUGIN_FILE' ) ) {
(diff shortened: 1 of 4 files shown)


includes/class-newspack-ads-scaip.php
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		$scaip_settings = array(
+			array(
+				'description' => __( 'Post ad inserter settings', 'newspack-ads' ),
+				'help'        => __( 'Super Cool Ad Inserter plugin options', 'newspack-ads' ),
+				'section'     => 'scaip',
+			),
+			array(
+				'description' => __( 'Number of blocks before first insertion', 'newspack-ads' ),
```
yogeshbeniwal, line 51: As now ads are inserted on count of paragraphs, better to change wordings to paragraph, instead of blocks.

miguelpeixe: This is a good point. Not sure if we should change it to "paragraph", though. The types of blocks are filterable, although they are only paragraphs by default.

If SCAIP upgrades to support the selection of allowed insertion blocks on the UI the "block" wording would make better sense.

cc @adekbadek

includes/class-newspack-ads-scaip.php
```diff
@@ -0,0 +1,97 @@
...
+
+		if ( ! defined( 'SCAIP_PLUGIN_FILE' ) ) {
+			return $settings_list;
+		}
+
+		$scaip_settings = array(
+			array(
+				'description' => __( 'Post ad inserter settings', 'newspack-ads' ),
+				'help'        => __( 'Super Cool Ad Inserter plugin options', 'newspack-ads' ),
+				'section'     => 'scaip',
```
adekbadek, line 48: This setting item does not have a `key`, which prompts PHP to log `Undefined index` notices in `Newspack_Ads_SCAIP::map_option_name` and `Newspack_Ads_Settings::get_setting_option_name`

adekbadek (changes requested):
Works as expected, but there should be no PHP notices logged.

includes/class-newspack-ads-scaip.php
miguelpeixe: Good catch! They section configuration should be valid without a `key` set. a816f15781c6368456ce5e4087888ec7f9c24c44 adds the necessary checks.

adekbadek (approved):

Issue #193, opened 2021-10-04
Option for ad label


yogeshbeniwal:
Ad Label clearly differentiate ads from main content. Recommended to provide an option to customise ad label for all ad placements.

kmwilkerson:
Add a text field to add label to ads. Example using "Advertisement":

![Screen Shot 2022-01-07 at 12 43 34 PM](https://user-images.githubusercontent.com/20666833/148591781-3cb8b9a0-c1e8-4e17-8752-ccd1e98905ce.png)

@miguelpeixe I do like [your idea ](https://github.com/Automattic/newspack-ads/pull/228#issuecomment-976733818)of making this a placement setting rather than a global setting since a label on certain positions would look odd (thinking of above/below header and sticky footer). This would mostly be used for in-content placements and blocks on the homepage.

yogeshbeniwal:
@miguelpeixe Thanks for quick PR. Should this Ad label works for In content after paragraph blocks also? No Label is shown while using legacy widgets for ad placement. Ok for sidebar and headers ads.

miguelpeixe:
The Ad labels should only work for ads placed through the global placements wizard. The disabling of legacy mode should allow for the label to be displayed in SCAIP ads.

yogeshbeniwal:
@miguelpeixe Thanks, it does. IMO from Ux/UI point of view can remove Label from Mobile Sticky Footer Ad unit

PR #228, opened 2021-11-18
feat: scaip placements

miguelpeixe:
image

Deprecate the use of SCAIP sidebars in favor of the new placements system. This PR creates placements based on the number of repetitions determined in SCAIP settings.

It is backwards compatible with existing Newspack Ad Unit widgets on SCAIP sidebars.

### How to test

1. On `master` branch, make sure you have Newspack Ad Unit widgets placed and with ad units selected on SCAIP sidebars
2. Switch to this branch, refresh your **Appearance -> Widgets** page and observe the sidebars are gone
3. Visit the Advertising -> Placements wizard and confirm the selected ad units are selected on the new UI
4. Visit a post and confirm that the ads also are also displayed as expected
5. Back to the Placements wizard, deactivate some placements and modify the ad unit for others and confirm the changes take place as expected


includes/class-newspack-ads-scaip.php
```diff
@@ -20,12 +20,28 @@ class Newspack_Ads_SCAIP {
...
 		add_filter( 'newspack_ads_setting_option_name', array( __CLASS__, 'map_option_name' ), 10, 2 );
+
+		// Placements hooks.
+		add_action( 'scaip_shortcode', [ __CLASS__, 'create_placement_action' ] );
+		add_filter( 'newspack_ads_placements', [ __CLASS__, 'add_placements' ] );
+
+		// Deprecate sidebar.
+		remove_action( 'scaip_shortcode', 'scaip_shortcode_do_sidebar' );
+		add_filter( 'scaip_disable_sidebars', '__return_true' );
+		add_filter( 'newspack_ads_placement_data', [ __CLASS__, 'get_ad_unit_from_widget' ], 10, 3 );
```
adekbadek, line 44: I think this is too heavy-handed. Some users might be used to managing their SCAIP settings in the widgets screen. The new UI should be an alternative one, not mutually exclusive with the default (widgets screen) UI.

adekbadek (changes requested):
> Visit the Advertising -> Placements wizard and confirm the selected ad units are selected on the new UI

This did work for me only if the Newspack Ads widgets was the first (or only) widget in the SCAIP widget area. 
Some users might have ad widgets prefixed or suffixed with something like "Article continues after the advertisment", which is just one of examples.

miguelpeixe:
> Some users might have ad widgets prefixed or suffixed with something like "Article continues after the advertisment", which is just one of examples.

I didn't know that was a thing. The sidebar description states `Please do not put more than one widget here.`

We can remove the sidebar shortcode action in case a selection has been made through our placement system. Still, I find having both methods available to be confusing. The user could visit the sidebar thinking they are changing the settings, whilst the actual setting is being overwritten by a previously selected setting on the placements system.

Would really like an opinion from @kmwilkerson on this.

includes/class-newspack-ads-scaip.php
miguelpeixe: > The new UI should be an alternative one, not mutually exclusive with the default (widgets screen) UI.

The issue is that it is not an alternative UI, it is a different system entirely for handling the SCAIP placement.

miguelpeixe:
Another alternative that keeps the proposed logic is to loop through the widgets on `get_ad_unit_from_widget()` instead of assuming there's just one. If it finds a Newspack Ad Unit widget, use it.

miguelpeixe:
287cba178c2cb403cc12f5f5dbf56bea29310edd updates `get_ad_unit_from_widget()` to loop through widgets and use the first found Newspack Ad Unit widget. This should handle cases in which there are multiple widgets.

The use of custom text widgets to prepend or append contextual information for the ads is not a good idea and should be discouraged. Other widgets' appearance is not tied to the rendering of creatives on the ad slot and can result in content out of context.

The support of custom text and labels can be discussed more universally from #193 and I think it's a good idea to think of it as a placement feature.

Note that CI tasks will fail due to #232

kmwilkerson:
> The new UI should be an alternative one, not mutually exclusive with the default (widgets screen) UI.

In my opinion, the new UI (within the ad wizard) should be the only place where Newspack users can make changes to SCAIP settings and placements. Having both options (widgets and ad wizard) feels confusing. I understand we need to keep the native widget area settings intact because of the wide use of the plugin outside of Newspack, but we should hide the widget options for Newspack customers and only allow options to exist in the ad wizard.

> Some users might have ad widgets prefixed or suffixed with something like "Article continues after the advertisment", which is just one of examples.

@miguelpeixe The solution you listed above sounds great in accommodating those that already have a text widget with an ad currently in place within a SCAIP placement. For any future needs/requests of this nature, we can handle case by case.

> Language used within the wizard

After polling the Newspack team, it's been decided to move away from "SCAIP" related language and will no longer be referenced within settings or placements in an effort to make the experience feel more native.

adekbadek (changes requested):
I'm a bit concerned with this feature so heavily interfering with another plugin, but maybe this discussion is purely theoretical and the feature serves 99% of use cases well. 
With this level of interaction, it might be a good idea to copy the SCAIP logic to this plugin.  
---
I've one change request – since it was decided to drop SCAIP references, the "Post ad inserter settings" [help text](https://github.com/Automattic/newspack-ads/blob/8997cfe9b2e4af28a7fef4fcd8bbd3abc03faba4/includes/class-newspack-ads-scaip.php#L63) should be updated.

yogeshbeniwal:
+1 for merging related SCAIP code in this plugin. 1 less plugin with that and development of SCAIP can be independent to newspack.

adekbadek (approved):

PR #234, opened 2021-11-24
update dependencies to fix build issues

### LINKED ISSUE #232: CI Tests failing
`cimg/php` images [recently stopped using PPA in favor of custom compiling PHP](https://github.com/CircleCI-Public/cimg-php/pull/127).

This breaks our tests because [we currently depend on the PPA to install the `soap` dependency](https://github.com/Automattic/newspack-ads/blob/master/.circleci/config.yml#L10), a requirement of `googleads-php-lib`.

We need to figure out a way to install the extension and the image doesn't appear to present a solution. We might need to use a different base image for our tasks. Ideas are welcome!

adekbadek:
We'd use a larger updates such as https://github.com/Automattic/newspack-newsletters/pull/625, but this is just to make CI jobs pass and unblock the other PRs.

Closes #232

## How to test

1. Make sure `npm build` & `npm start` work 
2. Do a basic smoke test, nothing should change in functionality


miguelpeixe (approved):
🎉

PR #251, opened 2021-12-03
feat: handle multiple gam accounts

### LINKED ISSUE #225: Handle multiple GAM accounts
If a user with multiple GAM instances available authorises this plugin, the particular GAM instance to use will be whichever happens to be the first one:

https://github.com/Automattic/newspack-ads/blob/05c05897b7d907132aad30b77f733e96fdd36914/includes/class-newspack-ads-gam.php#L137-L149

While not problematic for single-GAM-instance users, this poses issues when the user has more. There should be a GAM network code picker UI.

miguelpeixe:
Handles a user connection that has multiple GAM accounts and the ability to choose which should be used by Newspack Ads.

Closes #225 

### How to test

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1334
2. Make sure you are authenticated with Google OAuth (Connections wizard) and your account has at least 2 connected GAM networks.
3. Visit the Advertising wizard and click "Configure" on the Google Ad Manager card.
4. Confirm the new dropdown selector to choose between the available networks
5. Select a different network and make sure the ad units are updated accordingly
6. Refresh the page and confirm your selection persisted
7. Visit the Placements area and confirm you are able to use the ad units for the selected network

### DIFF
--- a/includes/class-newspack-ads-gam.php
+++ b/includes/class-newspack-ads-gam.php
@@ -130,6 +130,7 @@ class Newspack_Ads_GAM {
 	 * Get GAM networks the authenticated user has access to.
 	 *
 	 * @return Network[] Array of networks.
+	 * @throws \Exception If not able to fetch networks.
 	 */
 	private static function get_gam_networks() {
 		if ( empty( self::$networks ) ) {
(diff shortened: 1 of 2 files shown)


adekbadek (approved):

miguelpeixe:
fc643e1c259c41d98d924b80833cbf44a274715e ensures the returned value from `get_gam_available_networks()` is always an array.

adekbadek (approved):

PR #260, opened 2021-12-10
fix: allow prebid.js to load async

miguelpeixe:
The current implementation of the `prebid.js` script does not allow it to be loaded asynchronously. It requires an [extra webpack plugin](https://github.com/prebid/Prebid.js/blob/master/plugins/pbjsGlobals.js) on its build so it can correctly pull previously queued commands and inject itself to the window.

This PR implements the plugin, add the `async` attribute to the script and updates the prebid.js version

## How to test

Run `npm ci && npm run build` and make sure `DevTools` continue to display the logs as specified at #231.


dkoo (approved):
Works as described!

PR #274, opened 2022-01-03
fix(header-bidding): disable initial load for gtag

miguelpeixe:
Avoid double fetching when using header bidding by disabling initial load for GTag.

### Update

c21f1a2298c68e3a14fd07c88841a81a4c79c0fb initially proposed the use of the `newspack_ads_disable_gtag_initial_load` filter to determine the use of `disableInitialLoad()`. This is problematic because our `prebid_script()` method have particular variables that prevents it from being unnecessarily printed on the page. Not printing it means ads wouldn't load (unless a third-party calls for the `refresh()` method).

82ebc8718903856d2ac344d2cb10c763702c8b96 updates to call `disableInitialLoad()` directly on the `prebid_script()` method, which is a better practice in regard to how GPT and Prebid configuration scripts should interact.

### How to test

 - While on the master branch ensure you have experimental header bidding configured as described on #231 
 - Confirm you have a random Customer ID and a valid ad unit with a random Placement ID for header bidding
 - Visit the page where the ad unit is displayed with an additional `?googfc` parameter
 - Expand the displayed logs and notice the `Ad fetch count` is **2**
 - Check out this branch, refresh the page and confirm the count is **1**

### DIFF
--- a/includes/class-newspack-ads-bidding.php
+++ b/includes/class-newspack-ads-bidding.php
@@ -64,7 +64,6 @@ class Newspack_Ads_Bidding {
 
 		// Scripts setup.
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
-		add_filter( 'newspack_ads_disable_gtag_initial_load', [ __CLASS__, 'gtag_disable_initial_load' ], 100 );
 		add_filter( 'newspack_ads_gtag_ads_data', [ $this, 'add_gtag_ads_data' ] );
 		add_action( 'newspack_ads_gtag_before_script', [ $this, 'prebid_script' ], 10, 2 );
 	}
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works fine, but I couldn't replicate an Ad Fetch Count of more than 1 even on `master`. I followed the steps in #231 and enabled a single Above Header placement with a compatible unit with a test Media.net Placement ID. On both `master` and this branch, Ad Fetch Count in the console showed as `1`.

miguelpeixe:
Thanks @dkoo! Weird that you weren't able to test the duplicate fetch. If you add  `alert( 'prebid script' );` to line `280` of `class-newspack-ads-bidding.php` on `master`, does it execute the alert? I wonder if any missing configuration is preventing the script to run.

dkoo:
@miguelpeixe The alert does get executed with that line. Still seeing only one fetch on `master`, but this branch doesn't seem to cause any negative effects for me.

PR #277, opened 2022-01-05
fix: ad unit error handling

### LINKED ISSUE #275: Cannot create new ad unit with the same name as an existing ad unit
When we create a new ad unit via the Newspack Advertising wizard, it generates an ad unit code using `sanitize_title` and the human-readable name of the ad unit. However, if the resulting ad code is already being used by another ad unit (even one which has previously been deleted and thus isn't displayed at all in the wizard), the new ad unit won't be created—instead, it will fail silently without creating the new unit.

To replicate:

1. In Newspack > Advertising, add a new ad unit and note the name. Save and confirm that it appears in the list of ad units.
2. Add another ad unit and use the same name as the ad unit you created in step 1. Observe that the wizard doesn't display any error message, but the new ad unit isn't created.
3. Delete the first ad unit you created in step 1 and repeat step 2. Observe that the new ad unit still isn't created, even though the initial ad unit isn't shown in the wizard.

miguelpeixe:
GAM Ad unit creation does not have API error handling implemented. This PR implements error handling for API update/create methods with generic messaging and specific messaging for the reported "not unique" error.

Closes #275 

### How to test

1. While on the master branch, follow the steps in #275 to confirm the issue
2. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1424
3. Run `npm run build` for `newspack-plugin`
4. Attempt to create the ad unit with the same name again and confirm you are not redirected to the inventory and the appropriate error message appears on the top of the page.

### DIFF
--- a/includes/class-newspack-ads-gam.php
+++ b/includes/class-newspack-ads-gam.php
@@ -24,6 +24,7 @@ use Google\AdsApi\AdManager\v202111\AdUnitSize;
 use Google\AdsApi\AdManager\v202111\AdUnitTargetWindow;
 use Google\AdsApi\AdManager\v202111\EnvironmentType;
 use Google\AdsApi\AdManager\v202111\Size;
+use Google\AdsApi\AdManager\v202111\ApiException;
 
 require_once NEWSPACK_ADS_COMPOSER_ABSPATH . 'autoload.php';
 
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This looks 👍, but awaiting some feedback on https://github.com/Automattic/newspack-plugin/pull/1424.

PR #282, opened 2022-01-18
fix(placements): ensure active placement have ID

miguelpeixe:
Ensure the active placement have ID for the `get_placements_data_by_id()` method. The odd case of a placement ad unit being left blank after being filled can cause inconsistent placement data.

### How to test the PR

1. On the master branch select an ad unit for a global placement and save
2. Select blank (Select an Ad Unit) for the same placement
3. Visit a page where this placement should be displayed
4. Debug logs should display: `Notice: Undefined index: id in newspack-ads/includes/class-newspack-ads-placements.php on line 366`
5. Check out this branch, refresh the page and no notices should be thrown.

### DIFF
--- a/includes/class-newspack-ads-placements.php
+++ b/includes/class-newspack-ads-placements.php
@@ -347,6 +347,11 @@ class Newspack_Ads_Placements {
 				continue;
 			}
 
+			// Skip placements without ID.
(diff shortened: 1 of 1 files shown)


dkoo (approved):
💯

PR #290, opened 2022-01-24
feat: add OpenX, PubMatic and Sovrn to header bidding

### LINKED ISSUE #289: Add additional adapters to prebid
In an effort to expand the capability for publishers to add their own bidders with whom they have direct relationships, we'll continue to add more adapters over time. In our first batch of bidder additions, we'll add:

- [OpenX](https://docs.prebid.org/dev-docs/bidders/openx.html)
  - Required parameters: 
    - `platform` (pub ID)
    - `unit` (placement ID)
- [Pubmatic](https://docs.prebid.org/dev-docs/bidders/pubmatic.html)
  - Required parameters:
    - `publisherId` 
- [Sovrn](https://docs.prebid.org/dev-docs/bidders/sovrn.html)
  - Required parameters:
    - `tagid` (placement ID)

miguelpeixe:
Adds supports for OpenX, PubMatic, and Sovrn adapters to header bidding. OpenX and PubMatic requires a publisher-wide ID that should be configured on the Header BIdding section of **Newspack Ads -> Setings**. Sovrn does not require a publisher ID and should only be configured through the placement itself with a unit id.

image

image

This PR also implements a multi-select control for configuring which adapters should be enabled. This change should improve the current adapter availability check, which looks for its `active_key` to ensure that any required publisher-wide ID is configured.

**Important note**: all of the adapters modules are imported through the same `prebid.js` file. Ideally, we would have a separate js file for each adapter module so they are only loaded if enabled. This is not possible with how Prebid.js is currently built. Any attempt of a custom method for building it would end up with a lot of dependency duplication that the current tree-shook single file build does not have. That being said, we should still keep an eye on this file size as we implement more adapters.

Closes #289 

## How to test this PR

1. Check out this branch and make sure your`newspack-plugin` is up to date with master (for https://github.com/Automattic/newspack-plugin/pull/1475)
2. Update your `wp-config.php` to enable the experimental bidders: `define( 'NEWSPACK_ADS_EXPERIMENTAL_BIDDERS', true );`
3. Visit `Newspack Ads -> Settings` and on the Header Bidding section select all adapters, add random values to required IDs, enable **Debug mode**, and save
4. Visit the `Newspack Ads -> Placements` and add random placement IDs values to a valid placement with an ad unit
5. Visit the page where this placement is rendered and confirm the logs display connection attempt with all the bidders:

image

### Toggling adapters

1. Back to the settings, select only a couple of the adapters and save
2. Refresh the page with the recently edited placement and confirm the logs display connection attempt only for the selected bidders
3. Visit **Newspack Ads -> Placements**, edit a placement, and confirm you only see inputs for the selected bidders


includes/class-newspack-ads.php
```diff
@@ -73,6 +73,9 @@ private function includes() {
 		include_once NEWSPACK_ADS_ABSPATH . '/includes/class-newspack-ads-custom-label.php';
 		include_once NEWSPACK_ADS_ABSPATH . '/includes/class-newspack-ads-bidding.php';
```
adekbadek, line 74: Suggestion: move the `defined( 'NEWSPACK_ADS_EXPERIMENTAL_BIDDERS' )` check here, instead of repeating it in all four bidder files.

adekbadek (approved):
Works as described, left one suggestion.

PR #294, opened 2022-01-27
fix: ad label default value

miguelpeixe:
Changes the custom ad label default value from "Advertising" to "Advertisement".


dkoo (approved):
💯

PR #296, opened 2022-01-31
fix: dont query inactive ad units

### LINKED ISSUE #250: GAM ad unit sync - only port "active" ad units
When the user connects to GAM via OAuth, all ad units are ported into Newspack (active and inactive). To avoid unnecessary clutter in the UI, we should update to only bring in "active" ad units.

miguelpeixe:
Closes #250 

## How to test

1. Make sure you are connected with a GAM network
2. Visit the **Advertising -> Google Ad Manager -> Configure** page and your Ad Units dashboard at GAM
3. Make sure you have `inactive` ad units on your GAM dashboard
4. Confirm that on this branch they are no longer visible

### DIFF
--- a/includes/class-newspack-ads-gam.php
+++ b/includes/class-newspack-ads-gam.php
@@ -365,7 +365,7 @@ class Newspack_Ads_GAM {
 		// Get all non-archived ad units, unless ids are specified.
 		$statement_builder = new StatementBuilder();
 		if ( empty( $ids ) ) {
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Noting that with this change, [the toggle in Ad wizard UI](https://github.com/Automattic/newspack-plugin/blob/a869a8e8a14cb66e690dd6b9fd9db108345b288c/assets/wizards/advertising/views/ad-units/index.js#L175-L178) is no longer needed.

miguelpeixe:
Filed #304

Issue #295, opened 2022-01-29
Full width ad for one column template


yogeshbeniwal:
Full width ad for one column template will be a good option to have for ad blocks inserted within content. Not sure if this has to be taken here or SCAIP or Newspack theme repo.

kmwilkerson:
@yogeshbeniwal In the PR linked above, you'll be able to set a full-width background color to in-content ads within single column posts. Additionally, we'll also be adding a key-value pair to allow for "post template" targeting from GAM so that you can serve wider creative to single column posts specifically. Within the Newspack ads wizard, you can edit the ad units used for in-content placements to accept these wider standard sizes, or even custom sizes. Having the post template targeting option will allow you to use the same in-content ad units on all posts regardless of template, but only serve creative wider than 728px to single columns posts.

earcos:
Just wanted to bring attention to the fact that full width ads for one column template do not work on AMP :(

miguelpeixe:
@earcos can you provide a screenshot or describe how it's behaving?

earcos:
Sure. This is an example. This position is configured for 970x250 or 970x90 or 728x90. AMP version ONLY shows 728x90, regular version shows the 970x90 or 970x250 every time. HTML shows the 970x250 code, but with the hidden attribute.

https://hipertextual.com/2022/02/lobezno-wolverine-doctor-strange-2-giro-inesperado

<img width="1344" alt="CleanShot 2022-02-24 at 16 54 06@2x" src="https://user-images.githubusercontent.com/16387856/155559909-38642752-d511-4497-8a61-48934da593ae.png">

PR #297, opened 2022-01-31
feat: add template targeting key

miguelpeixe:
Add the `template` targeting key using [`get_page_template_slug()`](https://developer.wordpress.org/reference/functions/get_page_template_slug/) for more targeting coverage.

It should cover the targeting of wider creatives for #295 (https://github.com/Automattic/newspack-theme/pull/1699).

image

## How to test

1. Check out this branch and visit Advertising wizard
2. Notice the success message inside the GAM card confirming the creation of the new targeting key. (The wizard will always look for missing useful targeting keys on load and create them)
3. Create a post using the One-column template, visit the page and confirm the targeting is being configured by inspecting the searching for : `"template":`. You should find the `gpt` config like this:
image

### DIFF
--- a/includes/class-newspack-ads-gam.php
+++ b/includes/class-newspack-ads-gam.php
@@ -91,6 +91,7 @@ class Newspack_Ads_GAM {
 		'slug',
 		'category',
 		'post_type',
+		'template',
 	];
 
 	/**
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
👌

PR #301, opened 2022-01-31
hotfix: gam serialised method arg

miguelpeixe:
There is a logic error on how some of the `get_serialised_*` methods work. If the result of the query returns empty, it will always fall to return all of the entities.

This hotfix changes the default argument value to `null` so it returns an empty array as expected for empty results.


dkoo (approved):
Logic looks sound—any way to test this?

PR #302, opened 2022-02-01
hotfix: reindex filtered sizes

miguelpeixe:
The header bidding ad unit sizes should be reindexed after being filtered by `array_intersect()`. Otherwise, `wp_json_encode()` will generate an object keyed by the existing indexes instead of an array.

## How to test

1. On the master branch, create an ad unit with a couple of standard sizes and a non-standard one: `100x50`, `728x90`, `970x250`
2. Place this ad unit on a placement with a random Media.net placement ID
3. Visit the page, open DevTools and with the element inspector, search for: "mediaTypes"
4. Observe the `sizes` are an object keyed by its indexes
5. Switch to this branch, refresh the page and confirm the value of `sizes` is an array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.26.2-hotfix.1](https://github.com/Automattic/newspack-ads/compare/v1.26.1...v1.26.2-hotfix.1) (2022-02-01)
+
+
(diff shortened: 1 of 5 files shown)


dkoo (approved):
Works as described 👍

PR #281, opened 2022-01-18
fix: use responsive strategy for scaip placements on amp

miguelpeixe:
Flag SCAIP placement ad units as `responsive` so it can use the responsive strategy for AMP.

Closes #280 

## How to test

1. On master branch, create an ad unit with the sizes: `300x250`, `760x120`
2. Visit **Advertising -> Settings** and make sure your SCAIP settings are **non-legacy mode**
3. Visit **Advertising -> Placements** and set the SCAIP placements to use the created ad unit
4. Visit a post, inspect the SCAIP ad and observe it's a single `<amp-ad />` with both sizes
5. Switch to this branch, refresh and observe it's now one element per size

PR #306, opened 2022-02-03
Hotfix: responsive sidebar placements

miguelpeixe:
Just like #281, sidebar placements should also use the responsive strategy for AMP.

## How to test

Similar to #281: 

1. On the master branch, create an ad unit with the sizes: 300x250, 760x120
2. Visit Advertising -> Placements and set the "Article below content" to use the created ad unit
3. Visit a post, inspect the below content ad and observe it's a single <amp-ad /> with both sizes
4. Switch to this branch, refresh and observe it's now one element per size


dkoo (approved):
👍

PR #307, opened 2022-02-03
feat: apply custom label to ad unit block and widget

miguelpeixe:
Implements the custom label to ad units placed through the block or widget. Also adds a filter so that the list of elements that contain ads can be expanded by third parties.

### How to test

1. Set valid ad units to a global placement, as a post block, and as a widget
2. Confirm you have "Custom label" enabled at Newspack Ads -> Settings
3. Confirm all ads are rendered with the label with AMP on and off


dkoo (approved):
💯

PR #310, opened 2022-02-09
feat(header-bidding): multiple gam orders

miguelpeixe:
Implement multiple GAM orders support for header bidding in order to meet the different revenue shares each bidder agreement requires.

image

image

## How to test this PR

⚠️ The default price granularity makes the creation of orders take up to 15 minutes. For testing purposes, if you'd like to review with less granular buckets, add `price_granularity_key: 'low'` to line `119` of `src/wizard-settings/header-bidding-gam/order.js` so it generates a smaller set of line items.

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1517
2. Make sure your instance is properly connected to GAM
3. Visit Newspack Ads -> Settings
4. Configure all bidders with random values and make sure they are enabled
5. Click on Manage Orders or Create Order (the message varies according to existing orders)
6. If you have existing orders, observe the list of orders available and click "Edit" on any of them
7. If there are any found issues (missing local configuration), click on Fix issues
8. After the issues are fixed, change the shared revenue to any percentage and select a couple of bidders
9. Update the order and visit an order line item GAM's dashboard to confirm the percentage has been subtracted to the unit value and the targeting is set to the selected bidders:

Captura de Tela 2022-02-16 às 14 36 39

Captura de Tela 2022-02-16 às 14 37 12

### Multiple orders

image

1. Create a new order with a random revenue share and different bidders and make sure you have a duplicate bidder from the previous order
2. Observe the warning message stating that there are 2 orders targeting that bidder
3. Edit the order to remove the duplicate bidder and change the revenue share to another random value
4. Save the changes and confirm the warning message is gone and the updated values are matched on GAM's dashboard

### Archive order

1. While on the "Manage Orders" Modal, click to Archive an order
2. Confirm the order is archived on GAM
3. Archive all other orders and confirm your modal falls back to create a new one.


src/wizard-settings/header-bidding-gam/order-popover.js
```diff
@@ -0,0 +1,66 @@
...
+						{ __( 'Edit', 'newspack-ads' ) }
+					</MenuItem>
+					{ isDraft && (
+						<MenuItem
+							onClick={ () => {
+								if ( typeof onArchive === 'function' ) onArchive();
+							} }
+							className="newspack-button"
+						>
+							{ __( 'Archive', 'newspack-ads' ) }
```
dkoo, line 54: This is a fairly serious action, as it's irreversible on the GAM side. I'd suggest we pop up a browser warning asking if you're sure you want to archive this order. Source: I accidentally clicked "Archive" in the middle of testing before I got to that step. 😆

dkoo (changes requested):
This is a fairly big PR, so I'm going to do the review in a couple of rounds.

Round 1: I completed functional testing of all testing steps outlined in the PR description. All worked smoothly for me! I just have one small UI suggestion below.

I'll do another review round after I have a chance to read through the code changes in more depth.

src/wizard-settings/header-bidding-gam/order-popover.js
miguelpeixe: You actually can unarchive an order through GAM, but I agree this action deserves a confirmation! Done in 11eb58a4a1eefeea9cec594e546b4bd6dbf8ddab

miguelpeixe:
A significant change that you'll observe, which required a lot of refactoring, is changing the primary key and fetch of an order from its `price_granularity_key` to its ID. Since the initial approach was not considering multiple orders support, the order was attached to the determined price granularity so if, by any chance, the price granularity changed, the user would be prompted to create a new one.

src/wizard-settings/header-bidding-gam/order.js
```diff
@@ -0,0 +1,320 @@
...
+			if ( orderId || isLastAttempt ) {
+				// Unrecoverable error.
+				if ( typeof onUnrecoverable === 'function' ) await onUnrecoverable( pendingOrder, err );
+				setOrder( null );
+			} else {
+				// Make it fail unrecoverably if it fails on next attempt.
+				if ( pendingOrder?.order_id ) {
+					setLastAttempt( true );
+				}
+				setError( err );
```
dkoo, line 174: If creating an order results in an error message, if you go to create a new order the modal still shows the error. We should probably clear the message when starting to create a new order.

Screen Shot 2022-02-18 at 11 59 21 AM

dkoo (changes requested):
It's really coming together! Ran into some more issues particularly around the flow of fixing misconfigured orders and creating new ones.

includes/class-newspack-ads-gam.php
```diff
@@ -589,14 +587,17 @@ public static function get_serialised_creatives_by_advertiser( $advertiser_id )
...
-		$line_items            = [];
-		$service               = self::get_line_item_service();
-		$page_size             = StatementBuilder::SUGGESTED_PAGE_LIMIT;
-		$statement_builder     = ( new StatementBuilder() )->orderBy( 'id ASC' )->limit( $page_size );
+	private static function get_line_items( StatementBuilder $statement_builder = null ) {
+		$line_items        = [];
+		$service           = self::get_line_item_service();
+		$page_size         = StatementBuilder::SUGGESTED_PAGE_LIMIT;
+		$statement_builder = $statement_builder ?? new StatementBuilder();
+		$statement_builder->orderBy( 'id ASC' )->limit( $page_size );
```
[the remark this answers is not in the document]
miguelpeixe: If `$statement_builder` is `null`, line 599 will initialize a new and empty `Statement_Builder()`. This just means it doesn't have any queries so it'll return all results with the determined `orderBy` and `limit`.

includes/class-newspack-ads-gam.php
```diff
@@ -611,6 +612,18 @@ private static function get_line_items() {
...
 
+	/**
+	 * Get all line items given an order ID.
+	 *
+	 * @param int $order_id Order ID.
+	 *
+	 * @return LineItems[] Array of LineItems.
+	 */
+	public static function get_line_items_by_order_id( $order_id ) {
+		$statement_builder = ( new StatementBuilder() )->where( sprintf( 'orderId = %d', $order_id ) );
```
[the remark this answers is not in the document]
miguelpeixe: I don't think so, it'd be harder to maintain since we would have to clear the previously set clauses on every use. Instantiating on each use shouldn't have an impact on performance, I think.

includes/class-newspack-ads-bidding-gam.php
```diff
@@ -607,76 +750,110 @@ private static function archive_order( $price_granularity_key ) {
...
 
 		$config = get_option( self::get_option_name( 'config' ) );
 		if ( ! $config ) {
 			return new WP_Error( 'newspack_ads_bidding_gam_error', __( 'Missing config', 'newspack-ads' ) );
 		}
 
-		$price_granularity = Newspack_Ads_Bidding::get_price_granularity( $price_granularity_key );
+		if ( is_array( $order_id_or_config ) ) {
+			$order_config = $order_id_or_config;
+			$order_id     = $order_config['order_id'];
```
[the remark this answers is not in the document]
miguelpeixe: 2721377e8ed38876f0f8d4efed77b49164bd52cc ensures the incoming config has the proper attributes

includes/class-newspack-ads-bidding-gam.php
```diff
@@ -607,76 +750,110 @@ private static function archive_order( $price_granularity_key ) {
...
+		} else {
+			$order_id     = $order_id_or_config;
+			$order_config = self::get_order_local_config( $order_id );
+		}
+
+		if ( false === $order_config ) {
+			return new WP_Error( 'newspack_ads_bidding_gam_error', __( 'Missing order config', 'newspack-ads' ) );
+		}
+
+		$price_granularity = Newspack_Ads_Bidding::get_price_granularity( $order_config['price_granularity_key'] );
```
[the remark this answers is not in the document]
miguelpeixe: 2721377e8ed38876f0f8d4efed77b49164bd52cc ensures the incoming config has the proper attributes

src/wizard-settings/header-bidding-gam/order.js
```diff
@@ -0,0 +1,320 @@
...
+		( ! order?.line_item_ids?.length || totalBatches > ( order?.lica_batch_count || 0 ) );
+
+	const canSubmit = () =>
+		hasIssues() ||
+		! orderId ||
+		parseInt( config.revenueShare ) !== parseInt( order?.revenue_share ) ||
+		JSON.stringify( config.bidders ) !== JSON.stringify( order?.bidders );
+
+	const buttonText = () =>
+		orderId ? __( 'Update Order', 'newspack-ads' ) : __( 'Create Order', 'newspack-ads' );
```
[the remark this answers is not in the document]
miguelpeixe: The empty selection of bidders should actually be allowed and there was a bug in checking the selected bidders. bf9897ce6cd3d571a4588573e208a33428a89f0e fixes this check.

> A list of checkboxes might be more user-friendly, what do you think?

I agree but I think this should be a generic approach from the `<SelectControl multiple={ true } />` or another component that lives inside `newspack-components`.

src/wizard-settings/header-bidding-gam/order.js
miguelpeixe: You're right! 98f80314f5b2c0a67f598448f8f880c3c484365a removes the `error` prop that was being passed to the `<Order />` component.

includes/class-newspack-ads-gam.php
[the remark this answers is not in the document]
dkoo: Oh, I misread the logic. You're absolutely right.

includes/class-newspack-ads-bidding-gam.php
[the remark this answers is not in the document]
dkoo: Can confirm the warning is no longer thrown.

includes/class-newspack-ads-bidding-gam.php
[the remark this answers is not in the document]
dkoo: Can confirm the warning is no longer thrown.

src/wizard-settings/header-bidding-gam/index.js
```diff
@@ -68,151 +54,65 @@ const HeaderBiddingGAM = () => {
...
-		} finally {
-			if ( data?.count ) {
-				setOrderName( `Newspack Header Bidding v${ data.count + 1 }` );
-			} else {
-				setOrderName( 'Newspack Header Bidding' );
-			}
-			setInFlight( false );
-		}
+	const getActiveOrders = () => {
+		return orders?.filter( order => order.is_archived === false ) || [];
```
dkoo, line 58: Non-blocking: should we sort this list alphabetically? It seems to appear in a somewhat random order, neither alphabetical nor by order of creation.

Screen Shot 2022-02-18 at 1 54 19 PM

src/wizard-settings/header-bidding-gam/index.js
```diff
@@ -224,56 +124,125 @@ const HeaderBiddingGAM = () => {
...
+										order.bidders?.includes( bidderKey )
+									);
+									return (
+										bidderOrders.length > 1 && (
+											<Notice
+												key={ bidderKey }
+												isWarning
+												noticeText={ sprintf(
+													// Translators: %1 The name of the bidder. %2 The number of orders for the bidder.
+													__( '%1$s is being targeted by %2$d orders.', 'newspack-ads' ),
```
dkoo, line 218: Question, not issue: I'm not clear on the difference between an order that targets all bidders (all bidders selected) vs. one that targets "any" (no bidders selected). In the latter case should we also show this warning if there's another order targeting a bidder?

src/wizard-settings/header-bidding-gam/index.js
```diff
@@ -224,56 +124,125 @@ const HeaderBiddingGAM = () => {
...
-					</p>
-					{ error && error.data?.status !== '404' && (
-						<Notice isError noticeText={ error.message } />
+					{ activeOrders.length && (
+						<>
+							<Card noBorder>
+								{ activeOrders.map( order => (
+									<ActionCard
+										key={ order.id }
+										title={ order.name }
```
dkoo, line 157: Non-blocking: can this title be a link to the edit view?

dkoo (approved):
Didn't run into any problems after the latest changes. Approving with non-blocking suggestions!

src/wizard-settings/header-bidding-gam/index.js
```diff
@@ -224,56 +124,125 @@ const HeaderBiddingGAM = () => {
...
+				</Modal>
+			) }
+			{ editingOrder !== false && (
+				<Modal
+					title={
+						editingOrder ? __( 'Edit Order', 'newspack-ads' ) : __( 'Create Order', 'newspack-ads' )
+					}
+					onRequestClose={ () => ! inFlight && setEditingOrder( false ) }
+				>
 					{ unrecoverable && (
```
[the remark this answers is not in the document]
miguelpeixe: I was able to replicate the issue too, it was missing the reset of the config in the case of an unrecoverable error. Otherwise, the ID of the archived failed order is preserved and the button action attempts to update the archived order instead of creating a new one. ce8038b82a642888410e36eb842b19b152e4289d fixes this behavior.

src/wizard-settings/header-bidding-gam/index.js
miguelpeixe: Good call! Done in cfa93b7ed35c033989406b995330a4ac05f728be

src/wizard-settings/header-bidding-gam/index.js
miguelpeixe: Empty or all selected produce the same effect. Although we support creating orders with any bidder configuration, the warnings should help control potential conflicts with the setup. 1a84e881c9d99141a267da3cfd0ecdd1dfad4dac updates the warning to consider "empty" bidder configuration as "any".

src/wizard-settings/header-bidding-gam/index.js
miguelpeixe: It'd be nice, but `<ActionCard />` doesn't support `titleLink` as anything other than a `href` string. We don't expect users to edit orders too often, so I think it's fine.

src/wizard-settings/header-bidding-gam/index.js
dkoo: Confirmed, thanks for clarifying!

dkoo (approved):
Looks good! Thanks for the latest round of updates.

PR #313, opened 2022-02-18
feat: Ads providers and Broadstreet integration

### LINKED ISSUE #288: Integrate Broadstreet into Newspack Ads plugin
The goal is to as-natively-as-possible integrate Broadstreet into Newspack Ads as an ad provider. The [current Broadstreet WordPress plugin](https://github.com/broadstreetads/broadstreet-wp) should be able to offer a good starting point. 

Lazy loading (since it uses the GPT method) and header bidding features will not be applied to Broadstreet ads.

miguelpeixe:
Add support for multiple providers alongside GAM for integration with the placements system. Once #316 is implemented, it'll be the core method for displaying ads through Blocks and Widgets as well.

This PR also adds [Broadstreet](https://broadstreetads.com/) as a core provider, which is automatically enabled if [its plugin](https://wordpress.org/plugins/broadstreet/) is installed and active. It uses methods from the [plugin's `Broadstreet_Utility` class](https://github.com/broadstreetads/broadstreet-wp/blob/master/Broadstreet/Utility.php).

image

Some notes:
 - I have moved the GAM API class to the `providers/gam` directory for better organization
 - The existence of this new system calls for a refactor of the `blocks` and `model` classes, which mixes core methods with GAM-specific helpers. I plan to address this refactor in a different PR since it won't bring new features and its best to be reviewed separately.

Closes #288 

### How to test this PR

1. While on the master branch, make sure you have GAM connected and placements configured with valid ad units
2. Switch to this branch and https://github.com/Automattic/newspack-plugin/pull/1521
3. Confirm the previously configured placements continues to render as expected
4. Edit a placement and confirm the new select control with "Google Ad Manager" (If you don't have Broadstreet installed, it should only have GAM as an option)
5. Install Broadstreet and ping me for demo account access
6. Select a Broadstreet zone and confirm it renders with both AMP Plus on and off and AMP deactivated
7. Written tests should also pass


includes/class-newspack-ads-widget.php
```diff
@@ -36,6 +36,9 @@ public function widget( $args, $instance ) {
 		if ( ! newspack_ads_should_show_ads() || ! isset( $args['id'] ) || 0 === (int) $instance['selected_ad_unit'] ) {
 			return;
 		}
+		if ( ! Newspack_Ads_Providers::is_provider_active( 'gam' ) ) {
```
dkoo, line 39: Is there a use case for including Broadstreet ads in widgets?

src/blocks/ad-unit/view.php
```diff
@@ -16,6 +16,9 @@ function newspack_ads_render_block_ad_unit( $attributes ) {
 	if ( ! newspack_ads_should_show_ads() ) {
 		return '';
 	}
+	if ( ! Newspack_Ads_Providers::is_provider_active( 'gam' ) ) {
```
dkoo, line 19: Is there a use case for including Broadstreet ads in blocks?

dkoo (approved):
This works well and the code is a nice example of a clean, testable structure. No real blockers here, but I am wondering if there's a reason widgets and blocks are restricted to GAM units? I guess we'd need to add UI to select a provider to the Blocks and Themes repo to support additional providers, right?

miguelpeixe:
Thank you for reviewing!

Instead of adding the provider feature to blocks and widgets individually, #316 aims for a more complete approach of porting the entire "placement" configuration to work with these entities. The idea is that they'll work like a "dynamic" placement. This will allow not only for the custom providers to work but also header bidding and ideally everything placement-related.

PR #322, opened 2022-03-04
feat: ad unit block as a dynamic placement

miguelpeixe:
This PR refactors the **Ad Unit Block** to behave as a dynamic placement and should follow https://github.com/Automattic/newspack-plugin/pull/1550 to enable the use of block-based widgets.

Some adaptations were made to the placement system to support this type of "dynamic" placement. A dynamic placement is supposed to be registered and placed programmatically without being editable through the placements UI:

1. Placement registration now lives inside the method `Newspack_Ads_Placements::register_placement()` instead of through a filterable array. This allows for the required hooks to be applied at the registration level.
2. A new `show_ui` placement config prevents the placement to be displayed in the editor UI
3. Tweaks were made to allow a placement to be registered with `data` for programmatic provider, ad unit, and bidders settings

image

image

By using the block-based widget editor, the legacy Ad Unit widget should not receive any more updates and should eventually be deprecated.

#### Basic functionality

1. Checkout https://github.com/Automattic/newspack-plugin/pull/1550 to enable block-based widgets
2. While on master, create an Ad block with an ad unit on a page and as a widget
3. Check out this branch confirm the new block renders with the previously set ad units

#### Custom provider

1. Install the `Broadstreet` plugin
2. Edit the block to use a Broadstreet zone
3. Confirm you are able to see the ad size on the block editor and render the ad on a page/widget

#### Header bidding

1. Confirm you have GAM connected, AMP Plus enabled and `define( 'NEWSPACK_ADS_EXPERIMENTAL_BIDDERS', true );`
2. Enable header bidding, make sure you have enabled bidders with random config IDs and debug mode
3. Edit an Ad Unit Block with "Google Ad Manager" as provider and confirm you are able to add bidder IDs to the block
4. Visit the page with the ad and confirm you see the bid attempt logs in DevTools


thomasguillot:
Screenshot 2022-03-14 at 12 24 43

Why are we displaying all the different sizes? We should match the front-end and only display 1 size, depending on the viewport

thomasguillot:
Here are some recommendations:

Remove the settings from the sidebar and use the `Placeholder` to display them here instead.

1 Init

Once and Ad Unit is inserted (Save button), the user will see an ad placeholder. They can edit it (Edit button in the Toolbar) that would take them back to the `Placeholder` state.

2 Inserted

When unfocused, the ad placeholder should blend nicely with the rest of the content.

3 Unfocused

---

Bonus: Can we support "Background Color"? Is the user selects a color, we could add an `8px` padding to the ad for example.

miguelpeixe:
Nice! Thank you for this, @thomasguillot!

80b1d800b919d583f357c37164eb3e79ae342211 implements your suggestions.

### Multiple columns

image

image

I changed the toolbar button from the text "Edit" to use the ✏️  icon.

### One provider without header bidding

image

> We should match the front-end and only display 1 size, depending on the viewport

I don't think we should match the front-end in this case. Being provider-agnostic, the block doesn't know exactly how the ad is going to be rendered and each provider should have its own strategy to handle multiple sizes and viewports. I propose we use the maximum width and height from the size list to represent the ad slot.

In the example below you can see block is **728x250**, which is not an actual ad size, but pulled from **728**x90 and 300x**250**:

image

> Can we support "Background Color"? Is the user selects a color, we could add an 8px padding to the ad for example.

I'd suggest using the theme customizer for modifying ads background universally, but I don't see why we shouldn't support it on the block. 6f80aaf6f93080acad203db568d6bae12133a9c5 implements it. Here's a left-aligned ad block with custom background:

image

miguelpeixe:
I wonder, still, why does the edit/save method in the block view is better than using the sidebar controls, which feels like the official way of doing this.

miguelpeixe:
I wonder, still, why does the edit/save method in the block view is better than using the sidebar controls, which feels like the official way of doing this.

thomasguillot:
> I wonder, still, why does the edit/save method in the block view is better than using the sidebar controls, which feels like the official way of doing this.

Essential settings (settings required to make the block work) should live in the editor.
The sidebar should be reserved for "advanced" settings since the sidebar can be hidden.

thomasguillot:
> [80b1d80](https://github.com/Automattic/newspack-ads/commit/80b1d800b919d583f357c37164eb3e79ae342211) implements your suggestions.

Thank you so much for this @miguelpeixe.

I pushed some further changes https://github.com/Automattic/newspack-ads/pull/322/commits/5919ddd145b8b3b8c0f8c6a20116789c3821765d and https://github.com/Automattic/newspack-ads/pull/322/commits/2213f90a08d96a1b4d57f0bd47a6c8751352ac42

miguelpeixe:
> Essential settings (settings required to make the block work) should live in the editor.
> The sidebar should be reserved for "advanced" settings since the sidebar can be hidden.

Thank you for explaining and the final tweaks! 💯

src/blocks/ad-unit/index.js
```diff
@@ -17,42 +16,48 @@ import edit from './edit';
...
 	supports: {
 		html: false,
 		align: [ 'left', 'center', 'right', 'wide', 'full' ],
+		color: {
+			text: false,
+			background: true,
+		},
 	},
 	edit,
 	save: () => null, // to use view.php
```
dkoo, line 63: Non-blocking nit: we're not using `view.php` anymore, right?

dkoo (approved):
GAM and header bidding work great, and it's nice to see a preview of the size of the unit in the block editor.

Something I wanted to highlight when testing Broadstreet, just in case. I can't replicate this on a fresh test site, only on my existing local site, which had already had a Broadstreet account setup. When I switched to this branch, I saw the following PHP notice being thrown on both front-end and back-end:

```
PHP Notice:  Trying to get property 'zones' of non-object in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/broadstreet/Broadstreet/Vendor/Broadstreet.php on line 273
```

Then when I went to add a Broadstreet ad unit, the block preview in the editor showed nothing. (I could still select the empty block and click "edit" to change the unit.) The front-end also rendered no ad unit. After clearing, saving, then re-entering credentials in the Broadstreet admin, the above issues were resolved.

I'm not sure whether this will be a problem for any sites already using Broadstreet, but recording the workaround here just in case.

Approving despite the above because I can't replicate it again, even on the site where it first happened.

miguelpeixe:
I was not able to reproduce the warning but have some idea of the stack trace which leads me to believe that this is an issue that should be handled on the Broadstreet side.

The implementation of the Broadstreet provider uses the `Broadstreet_Utility::getZoneCache()` method to retrieve the available zones, which refreshes the cache through `Broadstreet_Utility::refreshZoneCache()` in case `$zones === FALSE`. This update calls the API method `getNetworkZones()`, which failed on your instance on what seems to be an unhandled error.

In case we're able to reproduce, we can handle this possible API error ourselves. Maybe an unhandled expired API token?

d711296253e79119f4972ca35b866e0e756c224c updates the comment on the `save()` method.

dkoo (approved):
That all makes sense—I agree that the stack trace indicates an unhandled exception in Broadstreet's code rather than something we can handle here.

PR #325, opened 2022-03-08
fix(header-bidding): ensure sizes for creative placeholders

miguelpeixe:
While attempting to create a header bidding GAM order without having any enabled bidder, it attempts to create without creative placeholders. This is because the creative placeholders are generated combining the accepted ad sizes from enabled bidders. This PR modifies the `get_all_sizes()` method to always return the default accepted ad sizes if no bidder sizes are found.

### How to test this PR

1. On a fresh instance, connect to GAM and enable header bidding through `define( 'NEWSPACK_ADS_EXPERIMENTAL_BIDDERS', true );`
2. While on master attempt to create a GAM order without having any bidder selected as enabled
3. Confirm the "required creative placeholder" error
4. Install this branch and confirm you are able to create the order without selecting a bidder

### DIFF
--- a/includes/class-newspack-ads-bidding.php
+++ b/includes/class-newspack-ads-bidding.php
@@ -288,6 +288,9 @@ class Newspack_Ads_Bidding {
 			),
 			SORT_REGULAR
 		);
+		if ( empty( $bidders_sizes ) ) {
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Seems to fix the error, and the `empty` checks look good to me! 👍

PR #329, opened 2022-03-14
refactor: unify responsive size map

miguelpeixe:
There currently are 2 different strategies to handle size mapping and responsive ads. Although we need both strategies to meet AMP and GPT requirements, the applied rules should be the same to avoid unexpected behavior.

This PR unifies the size mapping logic by implementing the `get_responsive_size_map()` method and simplifies how it's applied to both GPT and AMP.

### How to test this PR

There shouldn't be any noticeable change:

1. Check out this branch and make sure ad units with multiple sizes are displayed as expected for different viewports with AMP and AMP Plus
3. Written test should pass

### DIFF
--- a/includes/class-newspack-ads-blocks.php
+++ b/includes/class-newspack-ads-blocks.php
@@ -166,16 +166,18 @@ class Newspack_Ads_Blocks {
 						return $size[0] < 600;
 					}
 				);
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Didn't see any changes in ad behavior with this. Code looks simpler and more readable. 👍

PR #335, opened 2022-03-21
add dependabot and update newspack-scripts

adekbadek:
While debugging a `newspack-scripts` related issue, I've noticed that dependabot is not configured for this repository. It should be, as on the other repositories.


miguelpeixe (approved):
👌

PR #345, opened 2022-03-21
Update README.md

kmwilkerson:
Update the Newspack Ads Readme to more appropriately reflect full scope of plugin and recent updates


README.md
```diff
@@ -1,8 +1,18 @@
-# Newspack Google Ad Manager
+# Newspack Ad Manager
```
miguelpeixe, line 1:
```suggestion
# Newspack Ads
```

Shouldn't we preserve the plugin name at this title? We don't mention "Newspack Ad Manager" anywhere else.

kmwilkerson: Yes! Nice catch - thank you!

miguelpeixe (approved):
🙌

PR #346, opened 2022-03-21
feat: integrate ad refresh control plugin

miguelpeixe:
Integrate Ad Refresh Control plugin settings and allow its frontend scripts on AMP Plus.

Further details and test instructions are at https://github.com/Automattic/newspack-plugin/pull/1564


yogeshbeniwal:
Can this also take care of scenario where publishers are filling unsold inventory with Adsense. Adsense ads refresh is not allowed as per policy.

kmwilkerson:
@yogeshbeniwal If AdSense auto ads are used on the site, this plugin will not refresh them as it looks for GPT code only. If a publisher is filling unsold inventory with AdSense via GAM, it is the publisher's responsibility to block AdSense from refreshing by inputting the advertiser ID into the plugin settings.

dkoo (approved):
Tested locally and on a new test site—looks good to me!

yogeshbeniwal:
@kmwilkerson What would be the advertiser ID for adsense in this case, f08c47fec0942fa0 is not accepted by plugin settings.

kmwilkerson:
@yogeshbeniwal The advertiser ID for AdSense is `0`. I've added [an issue](https://github.com/Automattic/newspack-ads/issues/370) to automatically disable the refresh for this ID, but in the meantime you can input that value.

Issue #332, opened 2022-03-15
Reorganize code structure


miguelpeixe:
The introduction of providers (#313) calls for a reorganization of the codebase:

 - [x] Move GPT code away from the block class (after #322) (https://github.com/Automattic/newspack-ads/pull/348)
 - [x] Move `Newspack_Ads_Model` to GAM provider related directory (https://github.com/Automattic/newspack-ads/pull/348)
 - [x] Rename `Newspack_Ads_GAM` to `Newspack_Ads_GAM_API` (https://github.com/Automattic/newspack-ads/pull/348)
 - [x] Use `Newspack_Ads` namespace for simplification and better readability (https://github.com/Automattic/newspack-ads/pull/348)

PR #348, opened 2022-03-23
refactor: use namespaces

miguelpeixe:
This PR is part of #332. It implements the use of namespaces for better code readability and maintenance.

No changes should take place. To make sure all features continue to work as expected:

Check out https://github.com/Automattic/newspack-plugin/pull/1571

1. Tests should pass
2. Install SCAIP, Broadstreet and connect with GAM (through proxied OAuth or Service Account)
3. Enable header bidding through `define( 'NEWSPACK_ADS_EXPERIMENTAL_BIDDERS', true );`
4. Add a GAM and a Broadstreet unit to placements
5. Add mock IDs to header bidding with debug mode on
6. Add an ad unit block to a page
7. Add an ad unit widget to a sidebar
8. Modify settings through the wizard and make sure they persist after page refresh
9. Navigate through pages with ads and make sure the ads are displayed as expected the header bidding logs show bid attempts


includes/providers/gam/class-gam-api.php
```diff
@@ -5,6 +5,8 @@
  * @package Newspack
```
dkoo, line 5: There are three instances of the legacy `Newspack_Ads_Model` class name being called in this file, on lines 1438, 1448, and 1452. These cause a fatal error to occur when connecting to GAM for the first time, which results in an incomplete configuration.

includes/suppress-ads.php
```diff
@@ -5,6 +5,8 @@
  * @package Newspack
```
dkoo, line 5: There's an instance of the legacy `Newspack_Ads` class name instead of `Core` on line 33 here. This causes a fatal error to be thrown on the block editor:

```
PHP Fatal error:  Uncaught Error: Cannot use object of type WP_Error as array in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-plugin/includes/wizards/class-advertising-wizard.php:563
```

dkoo (changes requested):
Ran into a couple of fatal errors due to legacy class names that need to be updated. Details below.

Also, this doesn't seem to be related to this PR (it happens on `master` too), but while testing I noticed I'm unable to get an ad unit to display in the "Global: Below Header" placement. Tested and confirmed on both my local test site and a fresh JN test site.

includes/class-bidding.php
```diff
@@ -646,12 +653,12 @@ function( $bidder_id, $bidder ) {
...
 	public function register_api_endpoints() {
 		register_rest_route(
-			Newspack_Ads_Settings::API_NAMESPACE,
+			Settings::API_NAMESPACE,
 			'/bidders',
 			[
 				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ $this, 'api_get_bidders' ],
-				'permission_callback' => [ 'Newspack_Ads_Settings', 'api_permissions_check' ],
+				'permission_callback' => [ 'Newspack_Ads\Settings', 'api_permissions_check' ],
```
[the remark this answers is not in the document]
miguelpeixe: Sadly, you do have to use fully qualified classnames in callbacks. Check out this example from the PHP docs: https://www.php.net/manual/en/function.call-user-func.php#example-5081

includes/bidding-functions.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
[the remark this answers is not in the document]
miguelpeixe: Good call! c0c15265ff16a887bd8f5e251a20b2095cc8f3b7 restores the functions into the class file with simplified names, since they're now namespace there's no need for redundancy.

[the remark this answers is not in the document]
miguelpeixe: 8627cf4f4be6a598ec29ec40bcc08c749fbb0e34 fixes additional missing changes for `\Newspack_Ads\register_bidder`

includes/providers/gam/class-gam-api.php
miguelpeixe: Thank you for catching that! Fixed by 538ff78060837eb40ef7cb3f14a0fc8fac25c573

includes/suppress-ads.php
miguelpeixe: Fixed by dbe1710a54dfdf0f10af2abcf1cc90461c937509

miguelpeixe:
Thank you for the review, @dkoo! Along with your requests, 7d2803d699f0c861109a4b6efb27fb30802f037a also fixes some remaining required root namespaces.

miguelpeixe:
I was going to continue the refactor proposed in #332 in another PR, but the remaining changes seemed simple enough that could fit here.

c5544e4e20114e5b7be0b1e008b61f9691c9c5a6 creates `class-gam-scripts.php`, which houses GPT scripts that currently lives inside the `Blocks` class. Blocks are provider-agnostic and shouldn't have GPT related hooks on it.

It also moves the `Newspack_Ads\Model` to `Newspack_Ads\Providers\GAM_Model`. The current design of the plugin is more generic and all GAM related development should be scoped as such.

The main plugin has been updated to also match these changes: https://github.com/Automattic/newspack-plugin/pull/1571/commits/801f4af72efc5744bf3836fc6ab439dc68494550

miguelpeixe:
> I noticed I'm unable to get an ad unit to display in the "Global: Below Header" placement.

Nice catch! I pushed #349 with the fix 🙌

includes/class-bidding.php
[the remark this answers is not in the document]
dkoo: Whoops, you're right. My mistake!

dkoo:
> Thank you for the review, @dkoo! Along with your requests, https://github.com/Automattic/newspack-ads/commit/7d2803d699f0c861109a4b6efb27fb30802f037a also fixes some remaining required root namespaces.

Do we need to root-namespace _all_ WP functions, then? Thnking of `wp_parse_args` [here](https://github.com/Automattic/newspack-ads/pull/348/files#diff-89d47d0bcb43bf40e33a19f8e947c9942d5f8cc07882ea949a55eeecff35a2b7R85), `wp_enqueue_*` [here](https://github.com/Automattic/newspack-ads/pull/348/files#diff-89d47d0bcb43bf40e33a19f8e947c9942d5f8cc07882ea949a55eeecff35a2b7R163-R170) and `wp_json_encode` [here](https://github.com/Automattic/newspack-ads/pull/348/files#diff-feeba5f700d31013eea1367a103abccada717699bcc0961f90ec520061981e06R127-R129).

includes/providers/gam/class-gam-api.php
dkoo: Confirmed, thanks!

includes/suppress-ads.php
dkoo: Confirmed, thanks!

miguelpeixe:
Thank you for the review! 🙇

PR #349, opened 2022-03-24
fix(placements): add action hook

miguelpeixe:
The new placement registration method introduced by #322 also changed how the action hook for ad insertion is added. It assumes that the hook is exclusive for ad insertion, which may not be true. This PR removes the `has_action` check and always registers the ad insertion callback to the placement hook.

### How to test

1. While on the master branch confirm that no ad is rendered on the "Below header" placement
2. Check out this branch and confirm the placement rendering is working as expected


dkoo (approved):
This seems to fix the problem nicely.

PR #350, opened 2022-03-24
fix(gam): consider size map viewport count for responsive strategy

miguelpeixe:
The GAM ad unit code for AMP is currently verifying the size count to use the responsive strategy. This generates unnecessary calculation and additional markup for slots that only results in 1 viewport after the size map calculation.

This PR moves the size map calculation to the `get_ad_unit_amp_code()` method so it verifies the viewport count before deciding which strategy to use. It also makes the size map calculation more accurate, considering the filtered `$size` array for the sticky placement.

This change also fixes an issue where an AMP sticky ad would not properly render due to the additional markup generated by the responsive strategy, in case the ad unit is multi-sized.

It also refactors the `<amp-sticky-ad />` generation for better maintenance.

### How to test this PR

1. Make sure you have AMP Plus off, AMP plugin active, and check out this branch
2. Create an ad unit with sizes that would result in one viewport width: `300x250` and ~`290x200`~`300x200` - this should only affect multi-size sharing the same width.
3. Place this ad unit and confirm it's rendered with a singular multi-sized `<amp-ad />` tag and not contained by a `<div id="div-gpt-amp-*" />`
4. Also place an ad unit with multiple different sizes and confirm the responsive strategy continues to work as expected

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -166,8 +166,8 @@ final class GAM_Model {
 		$prepared_ad_unit['placement'] = $placement;
 		$prepared_ad_unit['context']   = $context;
 
-		$prepared_ad_unit['ad_code']     = self::get_code_for_ad_unit( $prepared_ad_unit, $unique_id );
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works as described! 👍

PR #351, opened 2022-03-24
fix(placements): remove default ad unit

miguelpeixe:
The suggested "default" ad units for the standard global placements are not being used and cause these placements to always render an empty div. This gets fixed if the user disables the placement or manually set a different value followed by an empty value.

This PR removes those default ad units since there's no current reason to keep them.

### How to test

1. While on master make sure you don't have the `_newspack_ads_placement_global_below_header` option. It will make the placements system use the `default_ad_unit`
2. Visit a page and confirm there's a `1.5rem` gap between the header and the content and there's an empty `<div class="newspack_global_ad global_below_header" />`
3. Check out this branch, refresh and confirm the gap and the `<div />` are gone


dkoo (approved):
Fixes the empty `div`s 🚀

PR #352, opened 2022-03-25
fix(gam): remove unused gpt script

miguelpeixe:
After #329 the size mapping part of the GPT script was simplified but some code from the previous approach remained. This PR removes the unused code and moves the comments to PHP to reduce printed bytes.

No changes should take place, AMP Plus and off-AMP ads should continue to render without errors.

c61185245ce40b57e0bb4b68ebeea66235b696cd (with improvements on 8ab8e5333f77753b5cc54c5d84f56e830b8c17cc) also isolates the common targeting key-val across ad units, which should, for most cases, be the entire targeting. This also aims to reduce printed bytes by not repeating the same targeting parameters for every ad unit on the page.

### DIFF
--- a/includes/providers/gam/class-gam-scripts.php
+++ b/includes/providers/gam/class-gam-scripts.php
@@ -132,7 +132,9 @@ final class GAM_Scripts {
 				for ( var container_id in all_ad_units ) {
 					var ad_unit = all_ad_units[ container_id ];
 
+					<?php
 					// Only set up ad units that are present on the page.
+					?>
 					if ( ! document.querySelector( '#' + container_id ) ) {
 						continue;
 					}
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works well in AMP Plus and non-AMP, and I didn't run into any issues. I wondered at first why we even ran `GAM_Model::get_ad_targeting` for each ad unit only to remove pretty much all of the targeting from the ad unit data, but I suppose it's because the `newspack_ads_ad_targeting` filter means we're not always sure what the common targeting key/value pairs will be since it allows other plugins to alter the targeting—is that right?

miguelpeixe:
Thanks, @dkoo!

>  it's because the `newspack_ads_ad_targeting` filter means we're not always sure what the common targeting key/value pairs will be since it allows other plugins to alter the targeting

That's exactly my thought, I figured although it's not a feature we currently use, it's a nice to have that's worth preserving.

PR #362, opened 2022-04-06
fix(blocks): ensure id uniqueness

miguelpeixe:
If the ID doesn't exist, the generated ID, formed by the page ID + the ad unit name, is not guaranteed to be unique. This hotfix ensures that the generated IDs are unique by including an index of block count.

This fixes issues for publishers that have been using multiple ad blocks on the same page using the same ad unit. Until the ad unit block is not manually edited, it won't have a random ID attached to it and the generated ID will take place.

### How to test

1. Create a new page with multiple Ad Unit blocks
2. Make sure the blocks have the same ad unit
3. Edit the page code to remove the `id` attribute of every ad unit block, so it forces the use of generated ID
4. While on the release branch, visit the page and confirm only the first ad renders
5. Check out this branch and confirm all ad units are rendering

Note: Being a single ad unit rendering multiple times, it must have multiple creatives attached to it.

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -16,6 +16,13 @@ use Newspack_Ads\Placements;
 final class Blocks {
 
 	/**
+	 * Amount of blocks rendered in the page.
+	 *
+	 * @var int
+	 */
+	private static $block_count = 0;
+
+	/**
 	 * Initialize blocks
 	 *
 	 * @return void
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Works fine. Why not use [`uniqid`](https://www.php.net/manual/en/function.uniqid.php), though?

PR #365, opened 2022-04-07
chore: move gpt initialization

miguelpeixe:
Some publishers push custom GPT scripts through plugins that either apply the script or that allow custom manual scripts to be printed on the page.

This change ensures that the `googletag.cmd` array is initialized as soon as possible on the page and prevents `googletag is not defined` errors.

### How to test

No changes should take place, with AMP Plus on make sure ads are still properly rendered on the page.


dkoo (approved):
I had seen some `googletag is not defined` errors in the past and wondered where those were coming from—thanks for the fix!

PR #366, opened 2022-04-07
feat(gam): size map filters and width threshold

### LINKED ISSUE #364: GAM size mapping rule should be filterable and improved
The current size mapping strategy uses a width difference ratio to determine which ad units should be rendered in the same viewport.

https://github.com/Automattic/newspack-ads/blob/3a86c51cc018bf6bb6571fa4be06979f7e88ce89/includes/providers/gam/class-gam-model.php#L604-L633

There should be a filter to allow publishers to change this ratio for [AMP](https://github.com/Automattic/newspack-ads/blob/3a86c51cc018bf6bb6571fa4be06979f7e88ce89/includes/providers/gam/class-gam-model.php#L543) and [GPT](https://github.com/Automattic/newspack-ads/blob/3a86c51cc018bf6bb6571fa4be06979f7e88ce89/includes/providers/gam/class-gam-scripts.php#L96) ads.

The default value should also be revisited to ensure that tablet ads can also be rendered in desktop viewports.

miguelpeixe:
Implements size map filters and a width threshold to allow tablet ads to share rotation with desktop ads regardless of their width ratio difference.

Closes #364 

### How to test

1. Create an ad unit with the following sizes: `320x180`, `320x480`, `640x360`, `960x540` and apply this ad unit to a global placement
2. With AMP Plus on, visit the page and inspect the generated GPT script (search for `all_ad_units`)
3. Confirm that the script contains the following `size_map` config parameter formed for the created ad unit:

```json
{
  320: [
    [320, 180],
    [320, 480],
  ],
  640: [
    [640, 360],
  ],
  960: [
    [640, 360],
    [960, 540],
  ],
}
```

On master, the viewport `960` doesn't include the `640x360` size.

4. Toggle AMP Plus off and refresh the page
5. Inspect the `<amp-ad />` element and confirm the grouping of tags also match the above, with a tag that looks like this: `<amp-ad width="960px" height="540px" data-multi-size="640x360" />`

You can also test disabling the threshold by adding the following filter:

```php
add_filter( 'newspack_ads_gam_size_map_width_threshold', '__return_false' );
```

Written tests should also pass.

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -540,7 +540,12 @@ final class GAM_Model {
 			);
 		}
 
-		$size_map = apply_filters( 'newspack_ads_multisize_ad_sizes', self::get_responsive_size_map( $sizes ), $ad_unit );
+		$size_map = self::get_ad_unit_size_map( $ad_unit, $sizes );
+
+		/**
+		 * Legacy filter for custom size map.
+		 */
+		$size_map = apply_filters( 'newspack_ads_multisize_ad_sizes', $size_map, $ad_unit );
 
 		// Do not use responsive strategy if the size map only results in one viewport or the ad unit is fluid.
 		if ( 1 < count( $size_map ) && false === $ad_unit['fluid'] ) {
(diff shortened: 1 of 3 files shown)


adekbadek (approved):
👌

PR #373, opened 2022-04-11
feat(integrations): disable refresh control for adsense

### LINKED ISSUE #370: Auto-exclude refresh for advertiser ID '0'
Since AdSense does not allow refreshing, let's set a default exclusion when advertiser ID is `0` and also add this note to the metafield desciption:

> AdSense does not allow their ads to be auto-refreshed. When Newspack detects that AdSense is the advertiser for any given impression, a refresh will not take place. 

Screen Shot 2022-04-11 at 9 54 30 AM

miguelpeixe:
Add `0` to the list of disabled advertisers for refresh control.

Closes #370

### How to test

1. Check out this branch, visit the Add-Ons page and make sure you have Ad Refresh Control enabled
2. Visit the Settings wizard and confirm you have no **Excluded Advertiser IDs**
3. Visit the front page, inspect the HTML and search for `advertiserIds`
4. Confirm the value is `[1]`. The values are sent as a PHP to JSON encoded map, so index `0` being of value `1` means disable `0`
5. Fill `1234` to the **Excluded Advertiser IDs** field and save
6. Confirm the value printed on the page is now `{ "1234": 1, "0": 1 }`

### DIFF
--- a/includes/integrations/class-ad-refresh-control.php
+++ b/includes/integrations/class-ad-refresh-control.php
@@ -32,6 +32,7 @@ class Ad_Refresh_Control {
 	 */
 	public static function init() {
 		\add_filter( 'newspack_amp_plus_sanitized', [ __CLASS__, 'allow_amp_plus' ], 10, 2 );
+		\add_filter( 'avc_advertiser_ids', [ __CLASS__, 'default_advertiser_ids' ] );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described 👍

Issue #333, opened 2022-03-15
Suppression API


miguelpeixe:
Management of suppression settings is currently performed at newspack-plugin through the `Configuration_Manager` class. It's best to move towards improving the plugin's API and deprecate the use of the `Configuration_Manager` class.

Suppression settings should be implemented using the settings api.

PR #374, opened 2022-04-12
chore: create suppression class and functions.php file

miguelpeixe:
This PR is a part of #333, creating a namespaced `Suppression` class in accordance with the plugin's overall structure and moves the `newspack_ads_should_show_ads()` public function to an unnamespaced file dedicated for such functions.

Next steps for future PR(s) should be:
 - Move `get_suppression_config()` and global suppression config away from `Providers\GAM_Model`
 - Create API for modifying global suppression config
 - Modify `newspack-plugin` to use this API

### How to test

No change should take place:

 - Change the global suppression config in the Newspack Ads Suppression wizard and confirm they are applied
 - Edit a post, toggle the "Don't show ads on this post or page" and confirm it is applied

### DIFF
--- a/includes/class-bidding.php
+++ b/includes/class-bidding.php
@@ -333,15 +333,15 @@ final class Bidding {
 
 				// Detect sizes supported by available bidders.
 				$sizes = array_intersect(
-					array_map( [ __CLASS__, 'get_size_string' ], $ad_data['sizes'] ),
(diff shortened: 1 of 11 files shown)


miguelpeixe:
Thank you @dkoo!

> Is it just for backwards compatibility with other plugins which may be calling that unnamespaced function?

It is! This function is used across other plugins and we've already suffered a bit from all the recent namespace changes and refactoring. It's important to keep those functions at least until we decide to properly deprecate them.

PR #375, opened 2022-04-12
feat(gam): strict container bounds for ad slot

miguelpeixe:
Prevents the rendering of slot creatives larger than the container offset width. Also puts it behind a filter so publishers can modify its behavior per ad unit.

It also proposes a bleed of 40px by default, so the centered creative can extrapolate at most 20 pixels on each side.

~~The `newspack_ad_global` was added a default `width: 100%;` to ensure that the container fills the available space. This shouldn't be an issue for regular units, only for the full-width template, which requires this PR: https://github.com/Automattic/newspack-theme/pull/1766~~

~~I'd also appreciate a confidence check on the use of the direct parent `offsetWidth` for this strategy. An alternative is to resolve the issue by case, for example, detect if the slot is within `entry-content`, use its width as the argument and ignore other cases.~~

Update: I've decided to tweak the approach to use declared "bounds containers" instead of the direct parent. It is a safer approach since floated blocks or other unusual approaches may get affected by the previous method, resulting in no ads being rendered.

As documented inline, the filterable list of bounds containers selectors can determine the available width for the slot. In case the slot is contained by one, the available width will be the bigger of the bounds container width or the direct parent offset width.

The proposed selectors are:

 - `.entry-content` - post content
 - `.widget-area` - sidebars
 - `.sidebar` - increase potential sidebar coverage
 
Ad slots outside of the above elements shouldn't have container bounds applied.

### How to test

Using an ad unit with 300x200, 970x90, and 728x90 sizes apply as a SCAIP position and with your browser viewport at approximately 1250px:

1. Visit a post with sidebar (default template) and confirm only the 728x90 creative is rendered
2. Visit a post with the full-width template and confirm it rotates between 728x90 and 970x90

On master, it renders the 970x90 creative on both templates because it fits on the viewport, but the container width can be smaller and extrapolates bounds:

| Master | This PR |
| --- | --- |
| image | image |

Change the viewport to approximately 900px and confirm the 300x200 creative is rendered for the post with the default template.

Add other units to other global placements and make sure the ads are rendering as expected.

### DIFF
--- a/includes/providers/gam/class-gam-scripts.php
+++ b/includes/providers/gam/class-gam-scripts.php
@@ -88,15 +88,26 @@ final class GAM_Scripts {
 				$sizes = array_values( $sizes );
 			}
 
+			/**
+			 * Filters whether the script should consider the container offset width
+			 * to filter out ad sizes that are too wide for the slot.
+			 *
+			 * @param bool  $strict_container_bounds Whether to consider the container offset width.
+			 * @param array $ad_unit                 Ad unit data.
+			 * @param array $sizes                   Ad unit sizes.
+			 */
+			$strict_container_bounds = apply_filters( 'newspack_ads_gam_strict_container_bounds', true, $ad_unit, $sizes );
+
 			$prepared_unit_data[ $container_id ] = [
-				'unique_id' => $unique_id,
-				'name'      => esc_attr( $ad_unit['name'] ),
-				'code'      => esc_attr( $ad_unit['code'] ),
-				'sizes'     => $sizes,
-				'fluid'     => (bool) $ad_unit['fluid'],
-				'targeting' => $ad_targeting,
-				'sticky'    => GAM_Model::is_sticky( $ad_unit ),
-				'size_map'  => GAM_Model::get_ad_unit_size_map( $ad_unit, $sizes ),
+				'unique_id'               => $unique_id,
+				'name'                    => esc_attr( $ad_unit['name'] ),
+				'code'                    => esc_attr( $ad_unit['code'] ),
+				'sizes'                   => $sizes,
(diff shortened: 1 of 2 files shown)


includes/providers/gam/class-gam-scripts.php
```diff
@@ -194,12 +228,48 @@ function( $a, $b ) {
...
+					 * Identify the bounds container for this slot and use its offset
+					 * width as bounds width.
+					 */
 					?>
+					var boundsWidth = 0;
+					findContainer:
+					for ( var i = 0; i < ad_unit['bounds_selectors'].length; i++ ) {
+						var selector = ad_unit['bounds_selectors'][ i ];
+						if ( typeof boundsContainers[ selector ] === 'undefined' ) {
+							boundsContainers[ selector ] = document.querySelectorAll( selector );
```
dkoo, line 243: Non-breaking: why is it necessary to cache the `boundsContainers` for each ad unit? It seems like you're only using it within this loop and never again?

dkoo (approved):
This seems to work really well, and the use of "bounds containers" is a really clever way to sidestep potential issues with floats/flex/other weird CSS that might affect the ad container.

Testing with some ad units that have both narrow and really wide units, I never saw a unit render a creative that exceeded the width of its bounds container.

includes/providers/gam/class-gam-scripts.php
miguelpeixe: This loop is executed for each ad unit on the page because we want to be able to have the `boundsContainers` filterable per ad unit.

Performance-wise, the majority of cases will not be affected by the decision of allowing different values per ad unit because of this caching strategy.

miguelpeixe:
Thanks, @dkoo! 🙇

PR #376, opened 2022-04-13
perf(providers): optimize data cache

miguelpeixe:
A small performance tweak for fetched providers' data.

Moves the cache to the `get_provider_data()` method level in order to increase its coverage across different methods.

### How to test

No changes should take place.

With GAM connected and Broadstreet installed and connected visit the Placements editor and confirm data from both providers are available for editing.

Provider methods tests passing should also confirm the expected data is being returned.

### DIFF
--- a/includes/class-providers.php
+++ b/includes/class-providers.php
@@ -33,11 +33,11 @@ final class Providers {
 	protected static $providers = [];
 
 	/**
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Didn't notice any ill effects from this change 👍

PR #398, opened 2022-05-03
feat(gam): targeting for tags and authors

### LINKED ISSUE #395: Add additional options to targeting
In #78 we added targeting for categories and different post types. It would be great if we could add tags and authors to the default keyvals as well.

miguelpeixe:
Add targeting support for tags and authors over GAM.

Closes #395 

### How to test this PR

1. Check out this branch and visit a post with multiple tags using the `?googfc` parameter
2. Make sure you have Co-Author Plus **disabled**, Inspect the targeting options, and observe the author and tags targeted
3. Enable Co-Author Plus, add a co-author to the post and confirm using the inspector that the authors' slugs are targeted

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -816,6 +816,27 @@ final class GAM_Model {
 				$targeting['category'] = array_map( 'sanitize_text_field', $categories );
 			}
 
+			// Add tags slugs to targeting.
+			$tags = wp_get_post_tags( get_the_ID(), [ 'fields' => 'slugs' ] );
+			if ( ! empty( $tags ) ) {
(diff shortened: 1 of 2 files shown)


yogeshbeniwal:
@miguelpeixe Quick note to also include scenario where tag base is changed using Permalink Settings.

miguelpeixe:
Do you mean the target `key` name for tags should follow the permalink base? I don't think it's the case. The key refers to the taxonomy name, which is not internally altered by defining a different permalink base.

yogeshbeniwal:
> The key refers to the taxonomy name, which is not internally altered by defining a different permalink base.

Thank you for clarification.

dkoo (approved):
Works with tags, authors, and guest authors. 👍

PR #399, opened 2022-05-04
fix(customizer): enqueue lodash dependency

miguelpeixe:
Lodash is not compiled to the final script so it must be enqueued as a dependency of the `customizer-control.js` script.

### How to test

Since the introduction of the block-based widget editor, lodash is enqueued by default to the customizer. To test this PR you must disable it by installing the [Classic Widgets](https://wordpress.org/plugins/classic-widgets/) plugin.

On the release branch, visit the customizer and confirm the `undefined` lodash error. Check out this branch, refresh and confirm the error is gone and you are able to edit and save Ad Placements.


dkoo (approved):
Fixes the issue!

PR #401, opened 2022-05-05
feat: constant to opt-out of AMP Plus

miguelpeixe:
Implements the `NEWSPACK_AMP_PLUS_ADS_DISABLED` flag to opt-out of AMP Plus.

Allows for AMP Plus features to be used on the Newspack ecosystem without necessarily activating it for Ads.

It also simplifies the method to remove `class_exists()` for `\Newspack\AMP_Enhancements`. The `method_exists()` function already performs the necessary checks to avoid missing class errors.

ef1441952228f763e9efe40839db62cda2514755 removes an unnecessary duplicate logic

### How to test

1. Check out this branch with AMP installed and the AMP Plus flag (`NEWSPACK_AMP_PLUS_ENABLED`) confirm you are still able to see Ads rendered through GPT (inspect the Ad element and confirm it is not a `<amp-ad />` element)
2. Add `define( 'NEWSPACK_AMP_PLUS_ADS_DISABLED', true );` to your `wp-config.php` and refresh the page
3. Confirm the Ads are now using the `<amp-ad />` tag and the GPT script is not rendered (inspect elements and search for `googletag.cmd`, it should not find results)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [1.33.0-hotfix.1](https://github.com/Automattic/newspack-ads/compare/v1.32.2...v1.33.0-hotfix.1) (2022-05-05)
+
+
+### Features
+
+* constant to opt-out of AMP Plus ([162d42c](https://github.com/Automattic/newspack-ads/commit/162d42c16956f20c9e1f214f34bb0754ae4dff94))
+
 ## [1.32.2](https://github.com/Automattic/newspack-ads/compare/v1.32.1...v1.32.2) (2022-05-04)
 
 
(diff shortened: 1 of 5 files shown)


adekbadek (approved):
Approving, but one small change would be welcome.

yogeshbeniwal:
@miguelpeixe May I know why we are adding another option for AMP Plus, is it resolving any known issue?

miguelpeixe:
@yogeshbeniwal this PR does not fix any known issue. Some sites may want to take advantage of AMP Plus for features that do not relate to Ads and want to keep using `<amp-ad />`. With `NEWSPACK_AMP_PLUS_ADS_DISABLED` you will ensure that AMP will be used for Ads regardless of AMP Plus.

PR #415, opened 2022-05-19
feat: activation hook defaults

### LINKED ISSUE #411: Default placement data
For every new publisher, or publishers just installing the Ads plugin for the first time, these default units will be attached to the appropriate placement automatically.

Example of the Below Header placement with a default “Newspack Below Header”:
image

This allows for appropriate and strategically positioned ads to be available effortlessly as soon as a GAM network code becomes available.

miguelpeixe:
Implements the following activation defaults, if the settings are not yet found on the database:

 - Default ad units assigned to their respective placements
 - SCAIP with the legacy option turned off (legacy should still be the default option for users that were already using Ads with SCAIP)
 - SCAIP number of repetitions set to 3 in order to use use all 3 in-article default ad units

Closes #411

### How to test

1. On a fresh install, check out this branch
2. Go through the onboarding wizard and activate the Ads plugin
3. Enable GAM and enter a legacy network code
4. Visit the Placements wizard and confirm the active placements have a default unit assigned to them
5. Visit the Add-Ons wizard and enable SCAIP
6. Visit the Settings wizard and confirm legacy mode is off and the number of repetitions is set to 3
7. VIsit the Placements again, edit the SCAIP placements and confirm the default units are assigned to them

### DIFF
--- a/includes/class-core.php
+++ b/includes/class-core.php
@@ -48,7 +48,7 @@ final class Core {
 	/**
 	 * Activation Hook
 	 */
-	public function activation_hook() {
+	public static function activation_hook() {
 		do_action( 'newspack_ads_activation_hook' );
 	}
 
--- a/includes/class-core.php
+++ b/includes/class-core.php
@@ -42,6 +42,14 @@ final class Core {
 		$this->includes();
 
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
+		register_activation_hook( NEWSPACK_ADS_PLUGIN_FILE, [ __CLASS__, 'activation_hook' ] );
+	}
+
+	/**
+	 * Activation Hook
+	 */
+	public function activation_hook() {
+		do_action( 'newspack_ads_activation_hook' );
 	}
 
 	/**
--- a/includes/integrations/class-scaip.php
+++ b/includes/integrations/class-scaip.php
@@ -46,6 +46,9 @@ final class SCAIP {
 		add_action( 'init', array( __CLASS__, 'register_placements' ) );
 		add_action( 'scaip_shortcode', [ __CLASS__, 'create_placement_action' ] );
 
+		// Setup Newspack Ads defaults on activation.
+		add_action( 'newspack_ads_activation_hook', array( __CLASS__, 'register_defaults' ) );
+
 		// Deprecate sidebar.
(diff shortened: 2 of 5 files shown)


includes/class-core.php
```diff
@@ -42,6 +42,14 @@ public function __construct() {
 		$this->includes();
 
 		add_action( 'wp_enqueue_scripts', [ __CLASS__, 'enqueue_scripts' ] );
+		register_activation_hook( NEWSPACK_ADS_PLUGIN_FILE, [ __CLASS__, 'activation_hook' ] );
+	}
+
+	/**
+	 * Activation Hook
+	 */
+	public function activation_hook() {
```
adekbadek, line 51:
```suggestion
	public static function activation_hook() {
```

Because otherwise: `Non-static method Newspack_Ads\Core::activation_hook() should not be called statically`.

includes/integrations/class-scaip.php
```diff
@@ -146,6 +149,25 @@ public static function create_placement_action( $atts ) {
 		do_action( sprintf( self::HOOK_NAME, $atts['number'] ), $atts );
 	}
 
+	/**
+	 * Setup Newspack Ads defaults on activation.
+	 */
+	public static function register_defaults() {
```
adekbadek, line 155: I can see the settings in the wizard, but the SCAIP placements are not inserted in an article. How to debug that?

image

adekbadek (changes requested):

includes/integrations/class-scaip.php
```diff
@@ -146,6 +149,25 @@ public static function create_placement_action( $atts ) {
...
+		$defaults = [
+			'repetitions'    => 3,
+			'legacy_widgets' => 0,
+		];
+		foreach ( $defaults as $key => $value ) {
+			$option_name = sprintf( '%s%s_%s', Settings::OPTION_NAME_PREFIX, 'scaip', $key );
+			if ( ! isset( self::OPTIONS_MAP[ $key ] ) ) {
+				$option_name = self::OPTIONS_MAP[ $key ];
```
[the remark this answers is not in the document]
miguelpeixe: FIxed by 9b6f788577e1d6853f02d5bac9d771579dcea8a9

adekbadek (approved):

PR #433, opened 2022-06-06
fix(gam): prioritize service account and other tweaks

miguelpeixe:
An available service account credential should be prioritized because it can be considered as a more stable connection than an OAuth token with an expiration date. This PR implements this change and removes the `can_use_service_account()` check because a service account connection method should always be available for a disconnected client.

The Google Ads SDK was also updated to the latest version.

### How to test this PR

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1694
2. Make sure your instance can use OAuth and connect with Google OAuth
3. Visit the GAM ad units wizard and confirm you see the synced ad units and the Service Account connection method at the bottom is **not** available
4. At the "Connections" page, disconnect from Google
5. Visit the wizard again and confirm you are able to use a Service Account connection
6. Upload a Service Account credential and confirm you are connected with synced ad units
7. Connect with Google OAuth again
8. Visit the ad units wizard and confirm you are able to remove the service account credential since it has been prioritized and is being used
9. Remove the credential and confirm you are still connected through Google OAuth

### DIFF
--- a/includes/providers/gam/class-gam-api.php
+++ b/includes/providers/gam/class-gam-api.php
@@ -172,7 +172,8 @@ final class GAM_API {
 	 * Get Service Account credentials.
 	 *
 	 * @param array $service_account_credentials_config Service Account Credentials.
-	 * @return object OAuth2 credentials.
+	 *
+	 * @return ServiceAccountCredentials|false OAuth2 credentials or false otherwise.
 	 */
 	private static function get_service_account_credentials( $service_account_credentials_config = false ) {
 		if ( false === $service_account_credentials_config ) {
@@ -274,7 +275,10 @@ final class GAM_API {
 	 * @return int GAM network code.
 	 */
 	private static function get_gam_network_code() {
-		return self::get_gam_network()->getNetworkCode();
+		if ( empty( self::$network_code ) ) {
+			self::get_gam_network();
+		}
+		return self::$network_code;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/providers/gam/class-gam-api.php
```diff
@@ -1394,23 +1386,18 @@ private static function get_connection_details() {
...
-			'connected'               => false,
-			'connection_mode'         => $connection_details['mode'],
-			'can_use_oauth'           => $can_use_oauth,
-			'can_use_service_account' => $can_use_service_account,
+		$connection_details = self::get_connection_details();
+		$can_use_oauth      = self::can_use_oauth();
+		$response           = [
+			'connected'       => false,
+			'connection_mode' => $connection_details['mode'],
+			'can_use_oauth'   => $can_use_oauth,
```
dkoo, line 1394: Non-blocking: from what I can tell, we're no longer using this variable anywhere, so we can probably remove it from this response as well as its associated method.

dkoo (approved):
Works well, and simplifies things a bit. We might simplify things even further as suggested below, but left it as non-blocking in case I'm missing a use case.

includes/providers/gam/class-gam-api.php
miguelpeixe: Good call! Done in e90c8194fc8c9e95d7d4b39631309b11e19dfb05, which also linted blank spaces due to my new editor settings 😬

PR #439, opened 2022-06-07
feat(placements): fixed height support

### LINKED ISSUE #409: Add min-height to GPT slots
To reduce CLS, GPT slots should have min-height set to the minimum available height configured for the slot according to the viewport.

miguelpeixe:
Implements optional fixed height for ad placements.

Editing a placement from the wizard:

image

Editing a block (dynamic placement):

image

Editing a placement from the customizer:

image

## Broadstreet

Since Broadstreet zones only support one size, a simple div with a fixed height will contain the ad.

## GAM

To handle a variety of cases, the handling of fixed height occurs in two steps:

 1. A server-side rendered `<style />` applies media queries with height rules following the _viewport -> sizes_ map returned from [`get_ad_unit_size_map()`](https://github.com/Automattic/newspack-ads/blob/c394b4b9460c49862e26fb067009f26d1772b6a4/includes/providers/gam/class-gam-model.php#L759)
 2. To handle the odd case of altered size map from bounds containers, the height is dynamically readjusted as soon as the frontend-calculated bounds containers data is available

The case of the second step affecting CLS is very unlikely, considering it'll most likely happen outside of the viewport. Ads within bounds containers are inside post content, sidebar, and columns. This step is only executed to guarantee that a prioritized smaller ad with a larger height (300x250) doesn't get contained with a smaller height from a larger ad (728x90) calculated by the initial `<style />`.

### Viewport

CLS is an issue within the viewport. Ads with a fixed height that don't render any creative outside of the viewport will be hidden from the page.

~~Mind that although we could consider applying that logic to also use the rendered creative to fit the height and remove potential blank spaces from the slot, this is problematic for 2 reasons:~~

~1. Refresh control can change the rendered creative and cause extra annoying CLS (during read-time)~
~2. Lazy load options will also affect the time the creative height is known, which is likely to be the time when the slot is in the viewport~

**UPDATE** 
After @kmwilkerson [contribution](https://github.com/Automattic/newspack-ads/pull/439#issuecomment-1152732343), the fixed height will only be applied immediately for ads above the fold. Other ads will have their height set to `auto` until the slot renders a creative, at which point the height and its size map rules will be determined to use the rendered creative dimension. The size map is updated in order to prevent refresh control to render a creative that breaks the bounds.

## Placement suppression per post

Fixed height is most likely to be used for placements above the fold, which are the "Above header" and "Below header" placements. It is also very common to only display ads for such placements on inner content, not the homepage. Ads without creatives for such cases would not be hidden because they are above the fold and doing so would defeat the purpose of a fixed height.

To better accommodate this common practice and prevent unnecessary blank spaces, this PR also implements suppression of ads for specific placements on a page or post:

image

This suppression occurs server-side and will not impact CLS for that page.

Closes #409 

## How to test this PR

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/1697
2. Confirm you have AMP off (or AMP Plus)
3. Configure valid Broadstreet and GAM
4. Set the "Above header" placement with a multi-sized ad unit: `728x90`, `970x90`, `970x250` and fixed height
5. Visit a page and confirm the slot is `250` in height, regardless of the rendered creative
6. Change your window size to below 970, refresh and confirm the height is now `90`

### Customizer

1. Enter the customizer, edit the same placement by removing the fixed height and saving
2. Visit the page and confirm that fixed height is not applied
3. Back to the customizer, apply again and confirm fixed height is restored

### Bounds containers

1. Draft a 3 column page (columns within > 300 and < 700 width)
4. In a column, add an Ad unit block with an ad unit with the following sizes: `300x250` and `728x90` with fixed height
5. Save and visit the page
6. Confirm the fixed height is set to `250`
7. Inspect the element and observe the `height: 90px;` rule being overwritten by the inline rule

### Hiding empty slots with fixed height

1. Editing the same page as above, add content above the columns to ensure that the ad unit does not render above the fold
2. Change the ad unit block to an ad unit that is guaranteed to not render a creative
3. Visit the page, scroll to the slot and confirm it's hidden

### Placement suppression

1. Editing the same page as above, suppress the "Above header" placement at the "Newspack Ads Settings" section
4. Visit the page and confirm the placement is not rendered
5. Visit another page and confirm the placement renders as expected

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -55,8 +55,8 @@ final class GAM_Model {
 	 * @return void
 	 */
 	public static function init() {
-		add_action( 'init', array( __CLASS__, 'register_ad_post_type' ) );
-		add_action( 'newspack_ads_activation_hook', array( __CLASS__, 'register_default_placements_units' ) );
+		add_action( 'init', [ __CLASS__, 'register_ad_post_type' ] );
+		add_action( 'newspack_ads_activation_hook', [ __CLASS__, 'register_default_placements_units' ] );
 		GAM_API::set_network_code( get_option( self::OPTION_NAME_GAM_NETWORK_CODE, null ) );
 	}
 
@@ -585,17 +585,17 @@ final class GAM_Model {
 	 * @param string $unique_id The unique ID for this ad displayment.
 	 */
 	public static function get_ad_unit_code( $ad_unit, $unique_id = '' ) {
-		$sizes        = $ad_unit['sizes'];
 		$code         = $ad_unit['code'];
 		$network_code = self::get_active_network_code();
 		$unique_id    = $unique_id ?? uniqid();
-		if ( ! is_array( $sizes ) ) {
-			$sizes = [];
+
+		if ( ! is_array( $ad_unit['sizes'] ) ) {
+			$ad_unit['sizes'] = [];
 		}
 
 		// Remove all ad sizes greater than 600px wide for sticky ads.
 		if ( self::is_sticky( $ad_unit ) ) {
-			$sizes = array_filter(
+			$ad_unit['sizes'] = array_filter(
 				$sizes,
 				function( $size ) {
 					return $size[0] < 600;
@@ -621,25 +621,26 @@ final class GAM_Model {
 	 * @param string $unique_id Optional pre-defined unique ID for this ad displayment.
 	 */
 	public static function get_ad_unit_amp_code( $ad_unit, $unique_id = '' ) {
-		$sizes        = $ad_unit['sizes'];
 		$code         = $ad_unit['code'];
 		$network_code = self::get_active_network_code();
 		$targeting    = self::get_ad_targeting( $ad_unit );
 		$unique_id    = $unique_id ?? uniqid();
 
-		if ( ! is_array( $sizes ) ) {
-			$sizes = [];
+		if ( ! is_array( $ad_unit['sizes'] ) ) {
+			$ad_unit['sizes'] = [];
 		}
 
 		// Remove all ad sizes greater than 600px wide for sticky ads.
 		if ( self::is_sticky( $ad_unit ) ) {
-			$sizes = array_filter(
-				$sizes,
+			$ad_unit['sizes'] = array_filter(
+				$ad_unit['sizes'],
 				function( $size ) {
 					return $size[0] < 600;
 				}
 			);
 		}
+
+		$sizes = $ad_unit['sizes'];
 
 		$size_map = self::get_ad_unit_size_map( $ad_unit, $sizes );
 
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -596,7 +596,7 @@ final class GAM_Model {
 		// Remove all ad sizes greater than 600px wide for sticky ads.
 		if ( self::is_sticky( $ad_unit ) ) {
 			$ad_unit['sizes'] = array_filter(
-				$$ad_unit['sizes'],
+				$ad_unit['sizes'],
 				function( $size ) {
 					return $size[0] < 600;
 				}
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -35,11 +35,11 @@ final class GAM_Model {
 	public static $custom_post_type = 'newspack_ad_codes';
 
 	/**
-	 * Array of all unique div IDs used for ads.
+	 * Associative array of all ad slots that will be rendered on the page.
 	 *
 	 * @var array
 	 */
-	public static $ad_ids = [];
+	public static $slots = [];
 
 	/**
 	 * Whether GAM units were already synced.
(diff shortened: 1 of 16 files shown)


kmwilkerson:
> Mind that although we could consider applying that logic to also use the rendered creative to fit the height and remove potential blank spaces from the slot, this is problematic for 2 reasons:
> 
> - Refresh control can change the rendered creative and cause extra annoying CLS (during read-time)

We could restrict the refresh to only request the same size that was initially rendered. Doing so would limit earning potential, but only very slightly.

> - Lazy load options will also affect the time the creative height is known, which is likely to be the time when the slot is in the viewport

Is it possible to set min height only for those units within the viewport? I realize that may be an over-complication, but if possible it may be a few things easier.  

> Placement suppression per post

Just to confirm, toggle off a placement would remove the define slot, not just hide it, correct?

yogeshbeniwal:
> this PR also implements suppression of ads for specific placements on a page or post:

We are already implementing best practices as default after thorough discussion. Providing option to suppress individual ad unit per post may lead to more support request, also how scalable/frequent it will be for editor to configure individual ads per post.

miguelpeixe:
> We could restrict the refresh to only request the same size that was initially rendered. Doing so would limit earning potential, but only very slightly.

Good call! I was able to achieve that by using `slot.defineSizeMapping()` with a new size mapping with only the rendered creative dimension, so only creatives matching that specific initial creative size are allowed on refresh. Implemented in 19d5ae86ab4d1c7c7622211091c5b1ff1da41706

> Is it possible to set min height only for those units within the viewport? I realize that may be an over-complication, but if possible it may be a few things easier.

That worked smoothly without affecting the CLS score. With units above the header, below the header, and several others below the fold using blocks I got the same `0.084` score. In comparison:

| | AMP | Fixed height | Non-AMP without fixed height |
| --- | --- | --- | --- |
| CLS Score | 0 | 0.083 | 0.372 |

Implemented  in 23ce2d41a26e270cf38a74a6d40fcc0fd0afd5d7

> Just to confirm, toggle off a placement would remove the define slot, not just hide it, correct?

Correct, the slot never reaches javascript or any front-end at all.

Thank you for your input @kmwilkerson!

miguelpeixe:
f69c28155528d3c2b345bb77978d9197c60d5d89 implements a slight change to our "viewport" logic. We can't just consider **in** viewport. Slots rendered on the past viewport will also affect CLS and should have their height fixed. Browsers now preserve scroll depth on simple refreshes.

miguelpeixe:
> We are already implementing best practices as default after thorough discussion. Providing option to suppress individual ad unit per post may lead to more support request. Also how scalable/frequent it will be for editor to suppress individual ads per post.

@yogeshbeniwal we currently don't have any appropriate solution to prevent a blank space to appear in case a publisher wants a good CLS score and doesn't want above-the-fold ads on a specific page. Unfortunately, no front-end calculation is soon enough in the render to avoid CLS.

includes/providers/gam/class-gam-model.php
```diff
@@ -583,25 +585,25 @@ public static function sanitize_sizes( $sizes ) {
...
+
+		if ( ! is_array( $ad_unit['sizes'] ) ) {
+			$ad_unit['sizes'] = [];
 		}
 
 		// Remove all ad sizes greater than 600px wide for sticky ads.
 		if ( self::is_sticky( $ad_unit ) ) {
-			$sizes = array_filter(
+			$ad_unit['sizes'] = array_filter(
 				$sizes,
```
adekbadek, line 599: Should be:

```suggestion
				$$ad_unit['sizes'],
```

includes/providers/gam/class-gam-scripts.php
```diff
@@ -255,80 +210,216 @@ function( $a, $b ) {
...
+						if ( ! ad_unit ) {
+							return;
+						}
+						<?php
+						/**
+						 * Lock fixed height ad unit size mapping to prevent larger
+						 * creatives to render on refresh control.
+						 */
+						?>
+						if ( ad_unit.fixed_height && container.parentNode.style.height === 'auto' ) {
```
adekbadek, line 352: In a particular setup on a local site, if the impression was not filled, the `size` of the `event` was `null`, which caused an error to be thrown. There should be a check for `event`'s `size`:

```suggestion
						if ( ad_unit.fixed_height && container.parentNode.style.height === 'auto' && event.size ) {
```

src/suppress-ads/index.js
```diff
@@ -16,34 +16,74 @@ import { withDispatch, withSelect } from '@wordpress/data';
...
 	render() {
-		const { newspack_ads_suppress_ads, updateSuppressAds } = this.props;
+		const placements = window.newspackAdsSuppressAds?.placements || {};
+		const {
+			newspack_ads_suppress_ads,
+			newspack_ads_suppress_ads_placements,
+			updateSuppressAds,
+			updateSuppressPlacements,
+		} = this.props;
+		console.log( newspack_ads_suppress_ads_placements );
```
adekbadek, line 26: ✂️
```suggestion
```

adekbadek (changes requested):
Left some minor tweaks, otherwise it's working well!

Is it correct to say that an ad with the fixed height setting on will display a blank space _only_ if the impression is not filled (obviously) _and_ the slot is visible in the view port on page load?

miguelpeixe:
> Is it correct to say that an ad with the fixed height setting on will display a blank space only if the impression is not filled (obviously) and the slot is visible in the view port on page load?

Correct!

Thank you for the review and catching those, @adekbadek 🙇

adekbadek (approved):

PR #456, opened 2022-06-30
fix: handling of default provider when GAM is not active

miguelpeixe:
The current implementation of the default provider assumes that GAM is always active and considered default. This causes unexpected behavior when Broadstreet is the only active provider and a placement does not include the `provider` attribute.

This PR fixes two issues:

 - Make sure GAM is active before using it as default. If not, the first active provider will be considered the default.
 - When placing an ad unit block with only 1 active provider, ensures the provider attribute is set to not rely on the default provider

## How to test

 - On the master branch, disable the GAM provider and enable Broadstreet
 - Attempt to place an ad unit block with a Broadstreet unit and confirm the ad unit does not render on the page
 - Also confirm that the "Save" button is greyed out because the provider is not set on the block
 - Check out this branch, refresh the page and confirm the ad renders
 - Edit the page again and confirm you are able to save and edit the unit
 - Enable GAM
 - Edit the page and confirm you are able to select and render GAM and Broadstreet units without issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.37.1-hotfix.1](https://github.com/Automattic/newspack-ads/compare/v1.37.0...v1.37.1-hotfix.1) (2022-06-30)
+
+
+### Bug Fixes
+
+* handling of default provider when GAM is not active ([d42141e](https://github.com/Automattic/newspack-ads/commit/d42141e36a860a57e0d3d7bd9d916ebecd0e9ac4))
+
 # [1.37.0](https://github.com/Automattic/newspack-ads/compare/v1.36.0...v1.37.0) (2022-06-27)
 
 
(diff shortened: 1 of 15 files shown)


dkoo (changes requested):
I'm not able to get the Placements UI to show me Broadstreet ad units in the select dropdown for specific units when Broadstreet is the only enabled provider.

Screen Shot 2022-06-30 at 12 22 41 PM

(Separately, I've noticed that the styles for the modal seem off, but this is probably from https://github.com/Automattic/newspack-plugin/pull/1676 and unrelated to the changes in this PR.)

miguelpeixe:
There was another bug while handling providers' units. Our own Broadstreet zones cache was using `is_admin()` check to update the cache or not. This check is `false` under the API request performed on the wizard so the data was never updated through that page.

93fba83f70490f3ad62291f390d619f6f1d24c3e implements the `$refresh` argument for a more clear management of this data.

dkoo (approved):
Works! Testing revealed another Broadstreet-related issue, but that can be handled separately.

miguelpeixe:
Thanks, @dkoo!

PR #478, opened 2022-08-09
fix(broadstreet): allow units without fixed sizes

miguelpeixe:
Properly handle Broadstreet zones without a fixed size. It's currently assumed that all zones will have a size and it causes zones without size to be rendered with `width` set to `0`.

### How to test

1. Setup Broadstreet
2. Create a zone without a fixed size
3. Attach an ad to this zone through a campaign
4. Set this zone to an ad placement
5. Visit the page and make sure the ad is rendered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.37.3-hotfix.1](https://github.com/Automattic/newspack-ads/compare/v1.37.2...v1.37.3-hotfix.1) (2022-08-09)
+
+
+### Bug Fixes
+
+* **broadstreet:** allow units without fixed sizes ([7985959](https://github.com/Automattic/newspack-ads/commit/79859594c704932ca7df3bb7c5339c6043905f08))
+
 ## [1.37.2](https://github.com/Automattic/newspack-ads/compare/v1.37.1...v1.37.2) (2022-07-26)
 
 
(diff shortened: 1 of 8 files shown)


dkoo (approved):
Approving because it avoids the PHP warnings and renders the ad, but noting here that when rendering in AMP, it results in an adorably tiny ad:

Screen Shot 2022-08-10 at 5 50 11 PM

We can address this is a separate PR if necessary.

miguelpeixe:
It's the consequence of not having a size on a flex box 😅 

Looking for alternatives now

miguelpeixe:
ef87fb184c355c362807c888147d9529688eea61 implements the required CSS adjustments so the Broadstreet ad renders as it should. You'll notice that this specific zone will render with a disproportionate height. That's because this sizeless zone is not meant to hold a static image and it's the same behavior as using a native Broadstreet shortcode.

I've also caught some other issues regarding sizeless units, which have been fixed on ef87fb184c355c362807c888147d9529688eea61.

To test those:

1. Edit a global placement through the customizer and notice the sizeless unit is shown with "Unknown size" text
2. Edit a page and place an Ad Unit block and observe the same design with text as above

dkoo (commented):
> You'll notice that this specific zone will render with a disproportionate height. That's because this sizeless zone is not meant to hold a static image and it's the same behavior as using a native Broadstreet shortcode.

So the really tall zone container is expected and okay? It did throw me off because it pushes the content all the way below the fold.

Confirmed the "Unknown size" behavior in Customizer and the block editor.

miguelpeixe:
> So the really tall zone container is expected and okay?

I wouldn't expect it either, but that's how the plugin's shortcode renders it as well.

You can see that this height is generated by Broadstreet itself, with a % `padding-top` that is unknown to us the logic behind it.

dkoo (approved):

PR #482, opened 2022-08-17
feat(suppression): expand options

miguelpeixe:


includes/class-suppression.php
```diff
@@ -84,60 +220,70 @@ function( $placement ) {
...
+			}
+			if ( ! empty( $config['categories'] ) && \has_category( $config['categories'], $post_id ) ) {
+				return false;
 			}
 		}
-
-		if ( true === $global_suppression_config['category_archive_pages'] ) {
-			if ( is_category() ) {
-				$should_show = false;
+		if ( \is_archive() ) {
```
dkoo, line 248: Testing this with the post type suppression, I find myself wondering why suppressing ads on a post type doesn't also suppress ads for that post type's archive pages, like with categories and tags. Also wonder why there isn't a global option to suppress ads on all post type archives, like with categories and tags, too.

dkoo (changes requested):
Small question/feature request below.

Also,

> This PR proposes a change in how we interpret those options. The selection of "all archive pages" for each taxonomy is preserved, but the selection of terms will suppress both archive and content that has the term.

This makes more logical sense to me than the current behavior. But it may result in more ads being suppressed than currently—do we need to inform publishers of this logic change?

includes/class-suppression.php
miguelpeixe: Oops! I added that but accidentally removed it while cleaning things up.

Fixed in c7b94b4703969bb84836eb74d353bec6b8eff814

dkoo: Confirmed!

dkoo (approved):

PR #508, opened 2022-09-14
test: update tests dependencies

leogermani:
This PR updates our phpunit tests dependencies to allow us to run them against the latest version of WordPress


dkoo (approved):
Tested by running `composer install`, reinstalling tests with the latest WP version, and running `./vendor/bin/phpunit` successfully. Thanks for the updates!

PR #521, opened 2022-09-23
fix(gam): identify insufficient permissions

miguelpeixe:
Adds "permission denied" to the supported types of generic error messages on GAM.

Ideally, we should be able to detect the authenticated session permissions beforehand so the user can address the issue immediately. I could not identify a solution [through Google's SDK](https://github.com/googleads/googleads-php-lib/blob/58.0.0/src/Google/AdsApi/AdManager/AdManagerSession.php) to implement such a feature.

This PR will just handle the error message if a privileged request is denied.

### How to test

1. Create a new service account with the role "Sales manager" (ping me if you'd like to use mine)
2. Connect to GAM using the service account JSON credentials
3. Attempt to edit/create an ad unit
4. Confirm you receive a comprehensive error message

### DIFF
--- a/includes/providers/gam/class-gam-api.php
+++ b/includes/providers/gam/class-gam-api.php
@@ -128,6 +128,9 @@ final class GAM_API {
 		if ( in_array( 'CommonError.CONCURRENT_MODIFICATION', $errors ) ) {
 			$error_message = __( 'Unexpected API error, please try again in 30 seconds.', 'newspack-ads' );
 		}
+		if ( in_array( 'PermissionError.PERMISSION_DENIED', $errors ) ) {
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
This is partially an issue with the Google API itself. The [docs](https://developers.google.com/ad-manager/api/authentication#scope) say

> Ad Manager has only one scope, shown below. 

So the OAuth process can grant access in this scope, but this does not mean the connected account has sufficient permissions to use the scope fully.

miguelpeixe:
> this does not mean the connected account has sufficient permissions to use the scope fully.

Yes, GAM has several different roles an account can be in.

PR #541, opened 2022-10-28
fix(placements): remove deprecated option handling

miguelpeixe:
This PR removes the handling of a deprecated option name. This has been up for almost 1 year and automatically migrating placement data if the deprecated option is found.

It's safe to remove this and stop querying the database with non-existing options.

## How to test

No change should take place, ad placements should continue to render as expected.


dkoo (approved):
Works without any issues!

PR #552, opened 2022-11-09
fix(deps): update @babel/runtime transitive dependency

adekbadek:
The JS package won't build with the version 7.16.7, see https://app.circleci.com/pipelines/github/Automattic/newspack-ads/1648/workflows/de1f974a-83ea-4a1b-872b-6291fabb1cfc/jobs/6062

This is not reproducible locally, so the real test will be if `build-distributable` job succeeds on `master` after this is merged.

Submitting as a `fix` commit, so a new alpha version may be released.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1859,11 +1859,11 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.16.7",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.16.7.tgz",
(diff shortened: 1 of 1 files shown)


dkoo:
Looks like this might be needed across some or all repos? https://app.circleci.com/jobs/github/Automattic/newspack-sponsors/1291?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

miguelpeixe (approved):

miguelpeixe:
Jumped the gun approving this PR! Ignore 😄

adekbadek:
This changes is not needed anymore – the JS files are built properly after changes in 49dd15fdaf285c53cef5630e561514bb6741d560 & 90bb588f705451e0ffe17b903423af1d469f04d1

PR #556, opened 2022-11-11
fix(gam): validate API session

miguelpeixe:
Implements #551 as a hotfix


miguelpeixe:
We'll need to test again and make sure it works on a JN to ensure the build process is not broken.

PR #557, opened 2022-11-11
fix(gam): ensure api requests are made with network code

miguelpeixe:
#551 surfaced an issue in which the network code for a GAM session may not always be available and they are required for most requests.

This PR ensures that if a network code is not yet set, it'll fetch from the top of the list to fulfill required requests.

The issue only occurs for fresh instances connecting GAM for the first time because the network code is not yet known.

### How to test

1. Make sure you're on the `release` branch
2. On a fresh install, upload a **valid** GAM service account credential
3. Confirm nothing happened (the request that validates the session couldn't fulfill because the network code is not yet known)
4. Check out this branch and confirm it's connected


leogermani (approved):
Works as described!

PR #566, opened 2022-11-24
feat(marketplace): product creation

miguelpeixe:
This PR implements the ad product as a custom WooCommerce product as part of the marketplace project. This and future PRs for this project should be merged into `epic/marketplace`.

image

The ad product will be initially defined with a CPD (cost per day) and its orders configured as sponsorships with a flat fee. The advertiser should be able to purchase a range of dates for the sponsorship to last (yet to be implemented).

The ad product configuration consists of:
 - Placements (required at least one)
 - Required sizes (required at least one)
 - Cost per day

The available placements are gathered from the active placements and their configuration. The selection of placements will determine the available sizes, which should all be available for the advertiser to upload creatives.

image

The proposed UI is a placeholder just for representing how the ad product is built. It will be later revised for a better experience.

## WooCommerce integration tests

The tests were updated to load WooCommerce for integration tests.

## How to test this PR

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2150
2. Confirm you have WooCommerce installed and active
3. Visit the Newspack -> Advertising wizard
4. Confirm you are connected to GAM and have active placements configured with ad units
5. Navigate to the new "Marketplace" tab
6. Confirm there's a button to "Create new product"
7. Click the button and confirm the modal is opened with the list of active placements for selection
8. Select different placements and confirm the unique sizes are displayed to select which are required
9. Select different sizes
10. Enter a CPD and save
11. Confirm the product is saved
12. Refresh the page and confirm the product persists
13. Edit the product values, save and confirm the values are updated
14. Visit the Products table, provided by WooCommerce, and confirm the links redirect to the wizard
15. Back to the wizard, click to delete the ad product
16. Confirm the product is deleted (not trashed)

PR #567, opened 2022-11-25
feat(marketplace): purchase sponsorship

miguelpeixe:
Implements basic ad order definitions and integration with GAM.

| WooCommerce Order | GAM Order Line Item |
| --- | --- |
| image | image |

An ad product order purchase should include:
 - From date
 - To date
 - Total days

We'll be using a temporary block called "Marketplace" for offering and purchasing ad products, which after product selection will redirect to checkout.

## How to test

1. Check out this branch and [newspack-plugin/epic/ads-marketplace](https://github.com/Automattic/newspack-plugin/tree/epic/ads-marketplace)
2. Make sure you are connected to a GAM account
3. If you don't have any Ad Products created, visit the Advertising -> Marketplace page and create some ad products (more details https://github.com/Automattic/newspack-ads/pull/566)
4. Edit a new page and add the "Marketplace" block, which renders only a placeholder component in the editor
5. Visit the page and confirm you see the products with date inputs:

image

6. Select a product, a valid date range (cannot be starting Today), and proceed to payment
7. Finish checkout and confirm the order is created in your local dashboard
8. Visit GAM, locate the created order and confirm the created line item data matches the ad product and order details:
   1. The creative placeholders matches the product
   2. The date range matches 12am of the "from" date and 11:59pm of the "to" date
   3. Total pricing should also be the same
9. The WooCommerce order should also have the values for the meta `newspack_ads_gam_network_code` and `newspack_ads_gam_order_id`, which will later be used for creating links for quick access to the GAM order

Added tests for validating cart item data should also pass.


dkoo (approved):
This is working as described! Since I know this feature set is still a WIP, I understand that the UI is in basic and subject to change. But I have a few non-blocking suggestions at this point. 

1. In the modal to create a new product in the Ads > Marketplace wizard, if the selected placement doesn't have any sizes to choose from, could we should a short message saying so? Otherwise the modal remains in an uneditable state with no indication of what's wrong.

Screen Shot 2023-06-12 at 1 36 58 PM

2. The formatting of the product summary in the checkout page looks a bit off. I'd expect the "To" and "Days" values to be on the same line as their labels, like with the "From" value:

Screen Shot 2023-06-12 at 2 11 48 PM

3. More non-blocking suggestions below:

miguelpeixe:
Thank you, @dkoo!

> In the modal to create a new product in the Ads > Marketplace wizard, if the selected placement doesn't have any sizes to choose from, could we should a short message saying so? Otherwise the modal remains in an uneditable state with no indication of what's wrong.

Absolutely! That's a great idea, although not possible in this PR because this lives in `newspack-plugin`. All of the UI should change after a more thorough design look, but I'll make sure to tweak this so it's also a good reference for important information that should remain in the design.

> The formatting of the product summary in the checkout page looks a bit off. I'd expect the "To" and "Days" values to be on the same line as their labels, like with the "From" value:

Looks like this is an issue with CSS coming from `newspack-theme`. If you open DevTools, select the `<dd />` element, and toggle off the `margin: 0;` from the `dl.variation dd, .wc-item-meta dd` selector it will fix the alignment. I'll do more testing to see how this impacts other checkouts and open a PR.

PR #586, opened 2023-01-19
Enable filtering of FE JS

adekbadek:
_Follow steps in https://github.com/Automattic/newspack-plugin/pull/2238_

### DIFF
--- a/includes/providers/gam/class-gam-scripts.php
+++ b/includes/providers/gam/class-gam-scripts.php
@@ -160,7 +160,7 @@ final class GAM_Scripts {
 		do_action( 'newspack_ads_gtag_before_script', $ad_config, $prepared_unit_data );
 
 		ob_start();
-		self::output_frontend_js( $ad_config, $prepared_unit_data, $common_targeting );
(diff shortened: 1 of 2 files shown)


includes/providers/gam/class-gam-scripts.php
```diff
@@ -158,11 +158,27 @@ function( $a, $b ) {
...
+	}
+
+	/**
+	 * Output frontend JS.
+	 *
+	 * @param array $ad_config Ad config.
+	 * @param array $prepared_unit_data Ad unit data.
+	 * @param array $common_targeting Common targeting.
+	 */
+	private static function output_frontend_js( $ad_config, $prepared_unit_data, $common_targeting ) {
```
miguelpeixe, line 176:
```suggestion
	/**
	 * Print GPT/GAM script.
	 *
	 * @param array $ad_config          Ad config.
	 * @param array $prepared_unit_data Ad unit data.
	 * @param array $common_targeting   Common targeting.
	 */
	private static function print_gpt_script( $ad_config, $prepared_unit_data, $common_targeting ) {
```

NIT: The parent method is called `insert_gpt_header_script`, we should follow a related naming.

PR #590, opened 2023-01-27
feat: "fixed height" a global setting with max threshold

miguelpeixe:
Changes how the "fixed height" system, implemented in #439, behaves. It's currently set as a per-placement setting and this PR turns it into a global setting with max height support:

image

If "max height" is enabled, any creative rendering above the defined threshold will cause a layout shift (CLS).

## How to test

1. Checkout this branch and https://github.com/Automattic/newspack-plugin/pull/2255
2. Visit **Advertising -> Settings** and confirm you see the new "Fixed height" section with the above defaults
3. Make sure you have GAM configured and set the Below Header and Footer placements with units containing the sizes: 728x90 970x90 970x250
4. Visit the home page and confirm the CLS occurring from 100 to 250 if the slot renders the 970x250 creative
5. Refresh until a creative with a height of 90px renders and confirm the slot height is at 100px without causing any CLS
6. Change the "Maximum fixed height" value to 150 and confirm the value is applied on the front-end
7. Disable the use of maximum fixed height, save, visit the front-end and confirm the slot is fixed with 250px of height and without any CLS
8. Visit the Customizer, edit the placement, and confirm the "Fixed height" setting is gone and you can save without issues
9. Visit "Advertising -> Placements", edit a placement, and also confirm the "Fixed height" setting is gone and you can save without issues
10. Edit a page, add an Ad Unit block, confirm the "Fixed height" setting is gone and you can save without issues

### DIFF
--- a/includes/class-fixed-height.php
+++ b/includes/class-fixed-height.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Newspack Ads Fixed Height Settings.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Ads;
+
+use Newspack_Ads\Core;
+use Newspack_Ads\Settings;
+
+/**
+ * Newspack Ads Fixed Height Class.
+ */
+final class Fixed_Height {
+
+	const SECTION = 'fixed_height';
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		add_action( 'newspack_ads_settings_list', [ __CLASS__, 'register_settings' ] );
+	}
+
+	/**
+	 * Register Fixed Height settings.
+	 *
+	 * @param array $settings_list List of settings.
+	 *
+	 * @return array Updated list of settings.
+	 */
+	public static function register_settings( $settings_list ) {
+		if ( Core::is_amp() ) {
+			return $settings_list;
+		}
+		return array_merge(
+			$settings_list,
+			[
+				[
+					'description' => __( 'Fixed height', 'newspack-ads' ),
+					'help'        => __( 'If enabled, ad slots will be rendered with a fixed height determined by the unit\'s sizes configuration.', 'newspack-ads' ),
+					'section'     => self::SECTION,
+					'key'         => 'active',
+					'type'        => 'boolean',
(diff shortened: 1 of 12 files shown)


includes/class-fixed-height.php
```diff
@@ -0,0 +1,71 @@
...
+					'description' => __( 'Fixed height', 'newspack-ads' ),
+					'help'        => __( 'If enabled, ad slots will be rendered with a fixed height determined by the unit\'s sizes configuration.', 'newspack-ads' ),
+					'section'     => self::SECTION,
+					'key'         => 'active',
+					'type'        => 'boolean',
+					'default'     => true,
+					'public'      => true,
+				],
+				[
+					'description' => __( 'Use maximum fixed height', 'newspack-ads' ),
```
adekbadek, line 50: I got a bit stumped by this label, thinking it's the same as the main setting. I think it'd be clearer like so:

```suggestion
					'description' => __( 'Set a maximum fixed height', 'newspack-ads' ),
```

yogeshbeniwal:
May I know that issue this change will fix? from CLS point of view only Above Header and Below Header is the issue, that can be solved by enabling current feature i.e. Use fixed height for individual ads.

Setting Maximum fixed height globally as one size fit for all is not a good approach as different ad location has different better performing ad sizes.

adekbadek:
This change enforces an opinion that all ad slots should have defined fixed height, with a sensible maximum. This mitigates most of the CLS, but the feature can be tweaked to anyone's liking – e.g. if you want 0 CLS (from ads at least), set a very high "maximum fixed height". If you prefer to optimise for as little empty space around ads as possible, set a lower value or disable the feature. 

Note that because the fixed-height feature will kick in for unfilled ads only for above-the-fold slots, this will in practice apply only to above- and below-header ads in most of the cases.

yogeshbeniwal:
> the fixed-height feature will kick in for unfilled ads only for above-the-fold slots

Thank you. This makes functionality clearer. Suggest to change descriptions to reflect 2 key things: unfilled ads + above-the-fold-slots only

PR #595, opened 2023-02-09
Add tag to GAM targeting when on a tag archive page

philipjohn:


miguelpeixe (approved):
Thanks, @philipjohn! It's working as described 👌

PR #610, opened 2023-03-03
fix(placements): suppress render for inactive providers

miguelpeixe:
The recent implementation of default fixed height for placements surfaced an issue with how the suppression for ads is applied when the provider is not active.

The suppression currently occurs only at the provider code level. The placement code is still applied and results in an empty div. This div is still processed by the fixed height CSS and results in a blank space.

This PR ensures the placement doesn't render any markup as well as the fixed height CSS.

## How to test

1. Confirm the issue in master by configuring placements with GAM active
2. Disable GAM, visit the frontend, and confirm the blank spaces
3. Checkout this branch, refresh, and confirm the units are not rendered
4. Enable GAM, refresh, and confirm the units are rendered again with proper fixed height

### DIFF
--- a/includes/class-placements.php
+++ b/includes/class-placements.php
@@ -719,6 +719,14 @@ final class Placements {
 			$placement_data = $placement['data'];
 		}
 
+		// Provider is not active.
+		if (
+			isset( $placement_data['provider'] ) &&
+			! Providers::is_provider_active( $placement_data['provider'] )
+		) {
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
This fixes the issue for me -- when GAM is off, I no longer see any gaps from the fixed-height ad spaces, but I do see the spaces when GAM is enabled again 🙌

miguelpeixe:
Thanks, @laurelfulford!

PR #647, opened 2023-04-20
feat(gam): support custom taxonomy targeting

miguelpeixe:
Implements ad targeting for any public taxonomy.

## How to test

1. Check out this branch and add `console.log(common_targeting);` to line 184 of `includes/providers/gam/class-gam-scripts.php`
2. Navigate to a post with categories and tags
3. Open DevTools and confirm the logged object has `category` and `tag` with the correct terms slugs
4. Edit a Woo product to have a product category
5. Navigate to that product and confirm `product_cat` contains the selected term slug

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -1059,16 +1059,24 @@ final class GAM_Model {
 				$targeting['template'] = sanitize_title( basename( $template_slug, '.php' ) );
 			}
 
-			// Add the category slugs to targeting.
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Approving because the following behavior is present in `master` too, but [this line](https://github.com/Automattic/newspack-ads/blob/master/includes/providers/gam/class-gam-scripts.php#L125) means that if there's only one ad unit on the page, no targeting info is logged. Is this intended behavior?

PR #648, opened 2023-04-24
Add .cache and .github to .distignore

philipjohn:
The release ZIP currently includes `.cache` and `.github` directories, which is probably not expected. This change excludes them both from the release ZIP.


miguelpeixe (approved):
Thank you, @philipjohn!

PR #694, opened 2023-07-21
feat(marketplace): orders api

miguelpeixe:
Implements ad product orders API.

This PR also reorganizes the code to implement an Api class and a separate Product class.

Instructions at https://github.com/Automattic/newspack-plugin/pull/2565


leogermani (approved):
Works well!

PR #721, opened 2023-09-07
fix: dont throw fatals if parentPath is null

claudiulodro:
Not sure how to force `getParentPath` to return `null` because [according to the docs it should always return an array (and an empty array if there is no parent)](https://developers.google.com/ad-manager/api/reference/v202305/InventoryService.AdUnit#parentpath). It can happen somehow though! If it does, we don't want fatal errors.


dkoo (approved):
LGTM! Holding for release until Monday.

PR #726, opened 2023-09-11
fix(gam): validate network code on session creation

miguelpeixe:
Validate the stored network code while creating the GAM API session. This change prevents user/credentials from getting stuck with an invalid network code when the connected account is updated.

It also adds cache deletion on option updates to prevent the 'notoptions' bug on memcache.

### How to test

1. While on the master branch, make sure you are connected to GAM
2. Manually change `_newspack_ads_service_google_ad_manager_network_code` and `_newspack_ads_gam_network_code` to an invalid value:
```
wp option update _newspack_ads_service_google_ad_manager_network_code "1"
wp option update _newspack_ads_gam_network_code "1"
```

3. Refresh the Advertising dashboard and confirm you get the "**The network code is invalid.**" error
4. Check out this branch, refresh, and confirm the error is gone and the network code is updated to match the network your account has access to.

### DIFF
--- a/includes/providers/gam/api/class-api.php
+++ b/includes/providers/gam/api/class-api.php
@@ -217,27 +217,6 @@ class Api {
 	}
 
 	/**
-	 * Whether the given network code is valid for the authenticated user.
-	 *
-	 * @param int $network_code Network code to validate.
-	 *
-	 * @return bool Whether the network code is valid.
-	 */
-	public function validate_network_code( $network_code ) {
-		$session = $this->get_session();
-		if ( \is_wp_error( $session ) ) {
-			return $session;
-		}
-		$networks = $this->get_networks( $session );
-		foreach ( $networks as $network ) {
-			if ( $network_code === $network->getNetworkCode() ) {
-				return true;
-			}
-		}
-		return false;
-	}
-
-	/**
 	 * Get GAM Connection User.
 	 *
 	 * @return User
--- a/includes/providers/gam/api/class-api.php
+++ b/includes/providers/gam/api/class-api.php
@@ -207,17 +207,34 @@ class Api {
 				'applicationName' => self::APP,
 			],
 		];
-		/** If a network code is not yet available, use first from list. */
-		if ( ! $this->network_code ) {
-			$session  = ( new AdManagerSessionBuilder() )->from( new Configuration( $config ) )->withOAuth2Credential( $this->credentials )->build();
-			$networks = $this->get_networks( $session );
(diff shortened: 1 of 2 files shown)


includes/providers/gam/api/class-api.php
```diff
@@ -207,17 +207,34 @@ public function get_session() {
...
+	}
+
+	/**
+	 * Whether the given network code is valid for the authenticated user.
+	 *
+	 * @param int $network_code Network code to validate.
+	 *
+	 * @return bool Whether the network code is valid.
+	 */
+	public function validate_network_code( $network_code ) {
```
dkoo, line 226: Am I missing something? I don't see this method being invoked anywhere.

dkoo (changes requested):
Functionally, it tests as describes. After updating my options to invalid values and refreshing the Ads wizard, it successfully reconnects to the GAM network. However, I don't see the new `validate_network_code` method being used anywhere, and I see the following upon refreshing the wizard:

In the debug.log, a PHP warning:

```
PHP Warning:  Network code is not specified. An API function will reject requests if it requires a network code. in /newspack-repos/newspack-ads/vendor/googleads/googleads-php-lib/src/Google/AdsApi/AdManager/AdManagerSessionBuilder.php on line 337
```

In the wizard, this warning:
Screenshot 2023-09-12 at 10 44 36 AM

The warning disappears after refreshing the page. I'm not entirely sure what it means. 

On the site front-end, GAM seems to be working okay: ad units render in the right placements, although in my tests the units weren't filling (probably not related to this change)

includes/providers/gam/api/class-api.php
miguelpeixe: Whoops! Thank you for catching that, it's residual code from a previous strategy I was going for. Removed in 7599f0741a1bf33bdf6963c9c4434d33b282cc05

miguelpeixe:
> In the debug.log, a PHP warning:

This warning is expected and comes from Google's sdk. On the session creation, we perform an API request without any network code in order to fetch the available networks. If that request was filled with a code that happens to be invalid it would error.

> In the wizard, this warning:

This warning is because of a mismatch detection of the legacy option with the authenticated network code option. Serves to alert publishers if they had a manually entered network code different than the one attached to the uploaded credential or oauth. You got it because in the tests both options changed.

dkoo (approved):
Thank you for explaining the warnings! Since they're expected from the artificial way we're changing the network code values, this looks good to me.

PR #729, opened 2023-09-15
fix(gam): default units array search matching

miguelpeixe:
If the found synced GAM ad unit is at index `0` it will be ignored by the current check and cause a faulty ad unit collection:

image

This PR fixes the index check by doing a strict comparison to `false`.


dkoo (approved):
Good catch!

PR #770, opened 2023-11-28
feat(gam): support "site" custom targeting key-val

miguelpeixe:
Implements "site" custom targeting key-val with type "PREDEFINED" and reportable type set to "CUSTOM_DIMENSION".

The `Targeting_Keys::update_default_targeting_keys()` method was also updated to use the `Targeting_Keys::create_targeting_key()` method instead of its own logic for better maintenance.

### Testing

1. If you have the `site` targeting key already created, visit your GAM Inventory -> Key-values, edit and delete each value, then delete the key (mind that GAM never deletes an entity, it's only archived)
3. Make sure you are connected to a GAM account in your Newspack instance
4. Check out this branch, visit Newspack -> Advertising, and confirm the targeting key is created:
image

5. Visit your GAM inventory and confirm the `site` is created with the previously available values plus your instance's URL

### DIFF
--- a/includes/providers/gam/api/class-targeting-keys.php
+++ b/includes/providers/gam/api/class-targeting-keys.php
@@ -32,13 +32,41 @@ final class Targeting_Keys extends Api_Object {
 		'category',
 		'post_type',
 		'template',
+		'site',
 	];
 
 	/**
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Works as described!

PR #777, opened 2023-12-07
fix(gam): parse targeting site url

miguelpeixe:
GAM limits a targeting key value to 40 characters. To reduce the chances of a site URL reaching the limit, this PR implements some parsing to remove the protocol, `www.` and trailing slash.

### How to test

Same as #770, but the value should now be `example.com` instead of `https://www.example.com/`


leogermani (approved):
Nit: the method is called `parse` but it's more `sanitize` or `prepare_for_gam`... parse is when we are reading something and extracting pieces from it

miguelpeixe:
Good call! I'll update to sanitize.

PR #816, opened 2024-03-12
fix: address dynamic prop warnings in gam api class

chickenn00dle:
Closes [0/1206638335395048/1206638704028951/f](https://app.asana.com/0/1206638335395048/1206638704028951/f)

This fixes some deprecated dynamic property warnings that were showing up in the logs when configuring GAM ad placements:
Screenshot 2024-03-12 at 13 00 13

### Testing instructions
1. If you haven't already, connect GAM via Newspack > Connections
Screenshot 2024-03-12 at 14 01 24
2. While monitoring `debug.log`, go to Advertising placement settings and start enabling some ad placements via Newspack > Advertising > Placements.
3. On `trunk` you will see some of the deprecated warnings in the image above. In this branch, you should not. Note that some of the Ad Manager dependency code will still trigger these warnings, but that is out of our control.

### DIFF
--- a/includes/providers/gam/api/class-api.php
+++ b/includes/providers/gam/api/class-api.php
@@ -73,6 +73,48 @@ class Api {
 	private $credentials = null;
 
 	/**
+	 * Advertisers
+	 *
+	 * @var Api\Advertisers
+	 */
+	public $advertisers = null;
+
+	/**
+	 * Creatives
+	 *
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Strange, I'm not seeing the warnings in debug.log, but I don't see any ill effects from this change, so approving!

PR #837, opened 2024-04-25
feat: add 1x1 to all size maps; deprecate AMP

dkoo:
Ensures that 1x1 is an allowed size for all ad units.

Also deprecates AMP support, which makes this PR a breaking change.

---
  - https://app.asana.com/0/0/1207127071414329


adekbadek (approved):

adekbadek (changes requested):
Approved too quickly, I failed to notice failing tests: https://app.circleci.com/pipelines/github/Automattic/newspack-ads/2438/workflows/e2b7fedd-6419-4ab4-95ef-66c4395ccad9/jobs/9863

dkoo:
Yeah, I saw the failing tests, too—not sure what that's about since it seems to have nothing to do with the changes. If I can get them to pass without further changes to non-test code, I'll merge

dkoo:
This ended up being a bit of a rabbit hole. The failing tests all had to do with AMP-related code. Since our platform no longer officially supports AMP, I ended up removing all AMP-related code from the plugin + tests. This makes this PR a `BREAKING CHANGE` and therefore not as urgent.

adekbadek (approved):
> Also deprecates AMP support, which makes this PR a breaking change.

In order to trigger a major release update, don't forget to include `BREAKING CHANGE` in the commit message body when squashing.

PR #859, opened 2024-06-13
fix: update deprecated v202305 GAM API to v202405

dkoo:
A recent release of the [`googleads/googleads-php-lib` library](https://github.com/googleads/googleads-php-lib/) deprecated the API version that we've been using in this plugin, resulting in connectivity issues and errors in the Newspack > Advertising dashboard. This PR updates the dependency and the API version being used.

### Testing

There should be no functional changes as a result of the update. All Newspack Ads features related to GAM should work as expected.

1. On `release`, on sites already connected to GAM, you might see an error in the Newspack > Advertising dashboard:

Screenshot 2024-06-13 at 12 43 36 PM

The error might also say one of the following:

- `Google Ad Manager Error: [ApiVersionError.UPDATE_TO_NEWER_VERSION @ version; trigger:'Version v202305 was deprecated and is now disabled. Please check the DoubleClick for Publishers blog for more information or contact us via the support forum.']`
- `[ApiVersionError.UPDATE_TO_NEWER_VERSION @ version; trigger:'Version v202305 was deprecated and is now disabled. Please check the DoubleClick for Publishers blog for more information or contact us via the support forum.']`

2. After checking out this branch, run `composer install` to fetch the updated library. The errors should no longer appear.

---
  - https://app.asana.com/0/0/1207564193443900

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 	},
 	"config": {
 		"platform": {
-			"php": "8.1.29"
+			"php": "8.1"
 		},
 		"allow-plugins": {
 			"composer/installers": true,
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
 		}
 	},
 	"require": {
-		"googleads/googleads-php-lib": "^61.0"
+		"googleads/googleads-php-lib": "^65.0"
 	},
 	"config": {
 		"platform": {
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 	},
 	"config": {
 		"platform": {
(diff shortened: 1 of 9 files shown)


claudiulodro (commented):
I wasn't able to fully test this yet, but it builds fine and doesn't immediately throw any errors. One thing you may want to do is make it a little easier to update the version for next time, because [GAM removes/deprecated/updates versions every 6 months](https://developers.google.com/ad-manager/api/deprecation)

adekbadek (approved):
Works as expected, the admin UI loads the ad units, and the frontend displays ads.

PR #876, opened 2024-07-29
feat: add Complianz integration

miguelpeixe:
Implements integration with the [Complianz plugin](https://br.wordpress.org/plugins/complianz-gdpr/) to determine whether there's consent to send reader data to GAM through the ad slot targeting strategy.

#### Test without the plugin

1. Check out this branch and make sure you don't have Complianz installed
2. In a fresh session, authenticate to a reader account and confirm the `reader_status` targeting key is added to the ad slot targeting. For that, visit a page with ads with the `?googfc` parameter and click "Open Targeting Info" of an ad slot.

#### With the plugin, but not configured

1. Install the Complianz plugin without configuring it
2. Repeat step number 2 of the previous test and confirm the targeting key is added

#### With Complianz configured 

1. Configure Complianz
2. In a fresh session, authenticate to a reader account and confirm the `reader_status` targeting key is not added
3. Close the GAM overlay and allow "Marketing" data through the consent banner preferences
4. Refresh the page with the `?googfc` parameter, recheck the targeting info and confirm the `reader_status` key is populated

---
  - https://app.asana.com/0/0/1207722207289374

### DIFF
--- a/includes/class-core.php
+++ b/includes/class-core.php
@@ -101,6 +101,7 @@ final class Core {
 		include_once NEWSPACK_ADS_ABSPATH . '/includes/class-placements.php';
 		include_once NEWSPACK_ADS_ABSPATH . '/includes/class-sidebar-placements.php';
 		include_once NEWSPACK_ADS_ABSPATH . '/includes/integrations/class-scaip.php';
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
🙌

PR #890, opened 2024-08-15
Alpha release Aug 15

adekbadek:
A new alpha release.


adekbadek:
https://github.com/Automattic/newspack-ads/pull/888 is already released, since it's a hotfix. So no new version was released here.

PR #894, opened 2024-08-22
feat(media-kit): add media kit (from publisher-media-kit plugin)

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Incorporates a Media Kit page functionality, adapted from the https://github.com/10up/publisher-media-kit/ plugin:

media-kit

1. If `publisher-media-kit` plugin is active, deactivate it and ensure the media kit page and `pmk-page` option are deleted
2. Use the WP CLI command `wp newspack-ads create-media-kit-page` to create the page, observe the command outputs the created post ID
3. Run the command again, observe it informs that the page has already been created, and outputs the post ID
4. Observe a media kit page draft was created, edit it 
5. Observe it has the content as in the image seen above
6. Publish the page, observe the interactive elements – the tabs in "Ad Specs" and "Our Rates" sections – are functioning properly

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you successfully ran tests with your changes locally?

---
  - https://app.asana.com/0/0/1207910087478332


leogermani:
It works well. I left the media-kit plugin active on purpose to see if there would be any conflict, looks ok. 

> If publisher-media-kit plugin is active, deactivate it and ensure the media kit page and pmk-page option are deleted

I had to manually delete the page.

Overall it looks good, my only question is that there is no way for an existing user of the plugin to start using this feature unless they deactivate and activate the plugin.

I'm not sure I have all the context, but is that ok? Don't we want to add a button somewhere to create the page?

adekbadek:
>  there is no way for an existing user of the plugin to start using this feature unless they deactivate and activate the plugin.
> I'm not sure I have all the context, but is that ok? Don't we want to add a button somewhere to create the page?

That's a great remark, indeed we should not require the plugin to be re-activated. However, in our case the migration will be one by one, assisted by our team, so we have no immediate need for a UI. Instead, I've added a CLI command in 7d259fcdc7467df33bdbd091acca504f2cb10ec8. I've updated the testing instructions.

leogermani (approved):

PR #931, opened 2025-01-08
fix: update CSS selector for tabs to work outside of media kit

laurelfulford:
The Tabs block was added to the Ads plugin as part of the Media Kit.

This PR changes a selector in the Media Kit styles, to make sure the tabs block looks ok outside of the Media Kit patterns.

### Steps to test

1. Add a tabs block to a regular post or page, and publish.
2. Note the appearance on the front-end -- it doesn't match the editor:

CleanShot 2025-01-08 at 10 50 48

CleanShot 2025-01-08 at 10 49 43

4. Apply this PR and run `npm run build`.
5. Confirm that the tabs block matches in the editor and on the front-end:

CleanShot 2025-01-08 at 10 51 12

### DIFF
--- a/src/blocks/tabs/editor.scss
+++ b/src/blocks/tabs/editor.scss
@@ -61,13 +61,8 @@
 				padding-top: 10px;
 			}
 
-			&:not(
-			:has(
-				.wp-block-paragraph[data-empty="true"]
-				+ .block-list-appender
-			)
-		) {
-			.block-list-appender {
+			&:not(:has(.wp-block-paragraph[data-empty="true"] + .block-list-appender)) {
+				.block-list-appender {
 					display: flex;
 					flex-direction: row-reverse;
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Looking better, but I notice there's a display bug when adding a new tab header. The header text is duplicated:

Screenshot 2025-01-09 at 4 01 29 PM

laurelfulford:
Thanks @dkoo! I noticed that, too, but I honestly wasn't sure if it was intentional 😂 

I've fixed the alignment -- can you please retest it as is, and inside of a group block with a background? The padding added to the tabs block in that situation caused the label to be pushed too far right as well, and this should now fix both!

CleanShot 2025-01-10 at 11 06 57

dkoo (approved):

Issue #170, opened 2021-09-02
Google MCM support with connected/synced GAM accounts


dkoo:
#169 adds support for using GAM accounts by defining network codes as a comma-delimited list, [as required by MCM](https://github.com/Automattic/newspack-ads/issues/157). This works for the "legacy" Ads UI that doesn't connect to a GAM account directly.

However, with #124 and now #163, it's possible to connect a Newspack site to a single GAM account using either an OAuth app or a Google service account. These implementations assume that each site can only be connected to a single GAM account at a time. We should update the implementation to support connections to multiple GAM accounts, for example if a service account is authorized to connect to more than GAM account for MCM. This will probably also require some Ad wizard UI updates to group synced ad units by connected GAM account, if the site is synced to more than one account.

PR #933, opened 2025-01-14
feat(gam): add filter for ad config data in gtag

miguelpeixe:
p1733840865403849-slack-C07UT1REAGJ

Adds a filter to GAM's script so the ad config data can be modified. This is mostly useful for Google MCM support, which allows a "parent" network code to manage the inventory for a "child" GAM account. See more on #170.

Eventually, we should implement the solution proposed in https://github.com/Automattic/newspack-ads/issues/170#issuecomment-925014706

### Testing

With GAM configured, check out this branch and confirm that the ads are rendered as expected. The addition of the filter shouldn't have any impact on GAM ad rendering.

### DIFF
--- a/includes/providers/gam/class-gam-scripts.php
+++ b/includes/providers/gam/class-gam-scripts.php
@@ -137,6 +137,13 @@ final class GAM_Scripts {
 		];
 
 		/**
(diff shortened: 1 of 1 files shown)


chickenn00dle (approved):
Looks good 👍

PR #947, opened 2025-05-30
fix(suppress-ads): check whether post type supports meta

miguelpeixe:
https://github.com/Automattic/newspack-plugin/pull/4000#issuecomment-2920758330

When a post type doesn't include support for `custom-fields`, the selector `select( 'core/editor' ).getEditedPostAttribute( 'meta' )` returns `undefined`. This needs to be handled in the editor plugin for ad suppression.

While at it, I also updated the `PluginDocumentSettingPanel` package dependency, which [moved to `@wordpress/editor` in WP 6.6](https://make.wordpress.org/core/2024/06/18/editor-unified-extensibility-apis-in-6-6/).

### Testing

1. Register a post type without support for `custom-fields`:

```php
register_post_type(
	'suppress-ads-test',
	[
		'label'        => 'Suppress Ads Test',
		'public'       => true,
		'supports'     => [ 'title', 'editor' ],
		'show_in_rest' => true,
	]
);
```

2. Navigate to `/wp-admin/post-new.php?post_type=suppress-ads-test`
3. Confirm you get the following error:

image

4. Checkout this branch, refresh the page, and confirm the error is gone
5. Edit a regular post a confirm the "Newspack Ads Settings" panel renders and behaves as expected

### DIFF
--- a/includes/class-suppression.php
+++ b/includes/class-suppression.php
@@ -201,7 +201,7 @@ final class Suppression {
 	 */
 	public static function enqueue_block_editor_assets() {
 		$post_type = \get_current_screen()->post_type;
-		if ( ! empty( $post_type ) && \is_post_type_viewable( $post_type ) ) {
+		if ( ! empty( $post_type ) && \is_post_type_viewable( $post_type ) && \post_type_supports( $post_type, 'custom-fields' ) ) {
 			\wp_enqueue_script( 'newspack-ads-suppress-ads', Core::plugin_url( 'dist/suppress-ads.js' ), [], NEWSPACK_ADS_VERSION, true );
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
This does a good job of suppressing the error message, but if a post type doesn't support custom fields, then the setting is meaningless, right? I still see the panel appearing in the editor sidebar, but updating it doesn't save any of the values I input. Maybe we should render `null` if `! meta`?

miguelpeixe:
Ah, good point! Added in 23f9074e1f82aeeee68f7ea79f2dcdea878eb08b

dkoo (approved):
:chefskiss:

PR #956, opened 2025-06-09
fix: avoid using get_option with random IDs [NPPM-2029]

naxoc:
The notoptions cache would sometimes fill up because of a option name with a uniqid string. This only gets the option if we know we can get it.

### DIFF
--- a/includes/class-placements.php
+++ b/includes/class-placements.php
@@ -275,7 +275,7 @@ final class Placements {
 		/**
 		 * Default placement data to return if not configured or stored yet.
 		 */
-		$default_data = wp_parse_args(
+		$data = wp_parse_args(
 			isset( $config['data'] ) ? $config['data'] : [],
 			[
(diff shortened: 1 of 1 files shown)


includes/class-placements.php
```diff
@@ -284,10 +284,13 @@ public static function get_placement_data( $placement_key, $config = array() ) {
 			]
 		);
 
-		$data = wp_parse_args(
-			json_decode( get_option( self::get_option_name( $placement_key ) ), true ) ?? [],
-			$default_data
-		);
+		// This check is on purpose. The show_ui key we look for is either be absent or set to true.
```
miguelpeixe, line 287:
```suggestion
		// This check is on purpose. The show_ui key we look for is either be absent or set to true.
		// Placements without admin UI will not have config stored in WP options.
```

miguelpeixe (approved):
Works! Added a comment suggestion, just to better clarify why we're skipping the `get_option()` for those placements.

PR #967, opened 2025-07-15
feat(gam): parent ad unit inventory

miguelpeixe:
Allow a site to select a parent ad unit to determine the site's ad unit inventory. If set, only the selected ad unit's children will be available to be used in placements.

image

### Testing

1. Before switching to this branch, make sure you have GAM properly connected and some placements with ad units
2. Checkout this branch and navigate to Advertising -> Providers and click on "Configure" GAM.
3. Confirm the new select field set to "None (all ad units will be used)"
4. Confirm all your placements continue to behave as before
5. Create a new ad unit and confirm it's created without a parent ad unit
6. Set an ad unit to a placement and confirm the placement uses the ad unit in the frontend
7. Back in the ad unit configuration page, change the parent ad unit
8. Confirm the selection persists and your inventory now shows the child units and the Newspack defaults
9. Create an ad unit and confirm in your GAM dashboard that it was created as a child ad unit
10. Visit Placements and confirm the available ad units for selection reflects the new inventory

### DIFF
--- a/includes/providers/gam/api/class-ad-units.php
+++ b/includes/providers/gam/api/class-ad-units.php
@@ -75,8 +75,8 @@ final class Ad_Units extends Api_Object {
 		);
 
 		$ad_units = [];
-				// Retrieve a small amount of items at a time, paging through until all items have been retrieved.
-				$total_result_set_size = 0;
+		// Retrieve a small amount of items at a time, paging through until all items have been retrieved.
+		$total_result_set_size = 0;
 		do {
 			$page = $inventory_service->getAdUnitsByStatement(
 				$statement_builder->toStatement()
--- a/includes/providers/gam/api/class-ad-units.php
+++ b/includes/providers/gam/api/class-ad-units.php
@@ -7,7 +7,6 @@
 
 namespace Newspack_Ads\Providers\GAM\Api;
 
-use Newspack_Ads\Providers\GAM\Api;
 use Newspack_Ads\Providers\GAM\Api\Api_Object;
 use Google\AdsApi\AdManager\Util\v202505\StatementBuilder;
 use Google\AdsApi\AdManager\v202505\ServiceFactory;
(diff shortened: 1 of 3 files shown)


rbcorrales (approved):

includes/providers/gam/api/class-ad-units.php
```diff
@@ -40,35 +39,79 @@ private function get_inventory_service() {
...
+	 *
+	 * @return array Array of serialzied AdUnits.
+	 */
+	public function get_parent_ad_units() {
+		$statement_builder = self::get_statement_builder();
+		$statement_builder->where( 'hasChildren = TRUE' );
+		$inventory_service = $this->get_inventory_service();
+		$page = $inventory_service->getAdUnitsByStatement(
+			$statement_builder->toStatement()
+		);
```
rbcorrales, line 75: Shouldn't this also be filtering by status?
If I deactivate a parent unit, I still see it listed for selection.

rbcorrales (changes requested):

includes/providers/gam/api/class-ad-units.php
miguelpeixe: Good catch! Fixed in 83a21e497de8ace8d1799eb2d0ea6f05d8742a4e

rbcorrales: Thanks for implementing this. However, if I had a parent selected, then deactivated in GAM, when I look at the config again, the parent shows "None" as selected, even though it's still being filtered by it because it's using the ID stored in the option without checking if active:
image

Ideally, it should check the status of the stored parent and update the content if not in the query of active parents, but I'm not sure if it would be a good practice to trigger an update operation from a frontend render. The other approach would be to just ignore the parent filter in this case (as if "None" were selected) and show all the children. WDYT?

miguelpeixe: You're right. I also think that updating on page load is not great because the ad unit might be temporarily inactive.

Pushed https://github.com/Automattic/newspack-plugin/pull/4084/commits/08bdadf1dbab2ef5d59d24ab94c9d5e8e941aa44 in https://github.com/Automattic/newspack-plugin/pull/4084 to implement a notice and an empty option instead:

image

WDYT?

rbcorrales: That also works, better to be explicit in this case 👍

rbcorrales (approved):

PR #974, opened 2025-07-23
feat(scaip):  add setting to filter after which blocks ads may be added

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

By default, Super Cool Ad Inserter inserts ads between paragraphs only. This change adds a setting to allow the SCAIP ads to be inserted between other kinds of blocks: image, heading, embed, on top of paragraph.

image

Closes NPPM-2102

1. Deactivate `super-cool-ad-inserter-plugin` plugin and visit Advertising → Settings
2. Observe no "SCAIP Block Settings" section
3. Activate `super-cool-ad-inserter-plugin` plugin and observe the "SCAIP Block Settings" section
4. Configure ads to display in `Post insertion #n` placements in Advertising → Placements
5. Visit a post, observe ads interspersed in content according to settings in Settings → Ad Inserter
6. Create a post with images only, observe no ads
7. In Advertising → Settings, adjust "SCAIP Block Settings" to include image blocks
8. Observe the ads inserted in the image-only post

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you successfully ran tests with your changes locally?


includes/class-scaip-block-settings.php
```diff
@@ -0,0 +1,100 @@
...
+					'description' => esc_html__( 'SCAIP Block Settings', 'newspack-ads' ),
+					'help'        => esc_html__( 'Configure which block types allow ad insertion within article content.' ),
+					'section'     => 'scaip_blocks',
+					'key'         => 'active',
+					'type'        => 'boolean',
+					'default'     => false,
+				],
+				[
+					'description' => esc_html__( 'Allowed Block Types', 'newspack-ads' ),
+					'help'        => esc_html__( 'Select which block types should allow ad insertion.' ),
```
miguelpeixe:
```suggestion
					'help'        => esc_html__( 'Select which block types should be included when calculating ad insertion. Warning: selected blocks are eligible to be followed by an ad.' ),
```

We should probably be clearer about what this means. Not sure if I did a good job with my suggestion, though 😄

miguelpeixe (changes requested):
The plugin has a SCAIP integration class with a settings section called "Post ad inserter settings" (`includes/integrations/class-scaip.php`). I think we should use that instead of a new one.

includes/class-scaip-block-settings.php
adekbadek: That's an odd way of putting it. This setting governs which block may be followed by an ad. "calculating ad insertion" sounds like it's about the percentage position of ad placement in a post. 

Updated to: "Select which block types can be followed by an ad.".

adekbadek:
@miguelpeixe – good call, updated in 9a1479e3304ab2d287149188fa2eedf1e430a9a3. I was looking for "SCAIP" mentions and the section uses "Post ad inserter settings" wording.

includes/class-scaip-block-settings.php
miguelpeixe: > "calculating ad insertion" sounds like it's about the percentage position of ad placement in a post.

But it is the case https://github.com/Automattic/super-cool-ad-inserter-plugin/blob/c11a669bf70c53442d2ee4b866fc27223451dd08/inc/scaip-shortcode-inserter.php#L111-L116

Note that `$block_index` only goes up if the block type matches the allowed list

adekbadek: I see, a different way of looking at it. But WDYT about "Select which block types can be followed by an ad".? IMO that's what it boils down to for the user. By default, only paragraphs can be followed by an ad. This PR adds ability to have more blocks followed by an ad.

miguelpeixe: Right, but on top of that, it also changes the calculation strategy. If I have an ad to be rendered **after 2 blocks** and the following article structure:

```html
<h2>Title</h2>
<p>Paragraph 1</p>
<p>Paragraph 2</p>
```

With default settings, it'll render here:

```html
<h2>Title</h2>
<p>Paragraph 1</p>
<p>Paragraph 2</p>
<!-- AD -->
```

If I include `Heading` in the options, it'll render here:

```html
<h2>Title</h2>
<p>Paragraph 1</p>
<!-- AD -->
<p>Paragraph 2</p>
```

I'm ok with phrasing, I just want to make sure the implications are clear 😄

miguelpeixe (approved):

includes/class-scaip-block-settings.php
adekbadek: Yeah, I'm worried this is too much of technical context :/

PR #975, opened 2025-07-23
feat(placements): new sticky and side rail

miguelpeixe:
Introduces new out-of-page side rail placements and extends the Sticky placement to no longer be just for mobile viewports. 

image

The side rail placements require custom display handling, as they are fixed and attached to the main content sides. The handling includes display animation and collision detection, preventing the ad from being rendered on top of full-width elements of the page:

### Regular header

https://github.com/user-attachments/assets/b54e7de0-90c0-4db2-9e4c-d395378b2ca0

### Sticky header

https://github.com/user-attachments/assets/dfa3e869-296b-4d2b-b5b1-fbd91651e6a6

## Testing

1. Add `define( 'NEWSPACK_ADS_SIDE_RAIL_PLACEMENTS', true );` to your `wp-config.php`
2. Navigate to Advertising -> Placements
3. Confirm you have the new side rail placements as shown in the image above
4. Confirm that the Sticky Footer placement text now includes the 768x90 size suggestion
5. Enable the Sticky Footer placement and select an ad unit with the suggested sizes (728x90, 320x50, 300x50)
6. Navigate to the frontend in regular desktop viewport and confirm the 728x90 creative renders in the slot
7. Change to mobile viewport, refresh the page, and confirm a 320x50 or 300x50 creative fills the slot
8. Back to the Placements configuration page, enable the side rails placements and select an ad unit with the 120x600 and 160x600 sizes
9. Navigate to the frontend, confirm that the ads render if there's available space
10. Tweak the window size and confirm the ad display animates in and out


kmwilkerson:
This is so much better than the native GAM behavior -- thank you!!!

includes/class-core.php
```diff
@@ -58,12 +58,21 @@ public static function enqueue_scripts() {
 		if ( ! newspack_ads_should_show_ads() ) {
 			return;
 		}
+		$asset_file = require_once dirname( NEWSPACK_ADS_PLUGIN_FILE ) . '/dist/frontend.asset.php';
```
chickenn00dle, line 61: Question: Does require_once always return the same value? Seems the result will depend on the file being required, which makes accessing version from the `$asset_file` variable feel a bit sketchy.

includes/providers/gam/class-gam-scripts.php
```diff
@@ -242,7 +242,7 @@ function inOrPastViewport( element ) {
 						availableWidth = Math.max( boundsWidth, containerWidth ) + parseInt( ad_unit['bounds_bleed'] );
 						for ( viewportWidth in ad_unit['size_map'] ) {
 							var width = parseInt( viewportWidth );
-							if ( shouldUseBounds && width > availableWidth ) {
```
chickenn00dle, line 245: Question: Is removing this `shouldUseBounds` intentional? Looks like this was checking the width of the container or something?

chickenn00dle (changes requested):
This is mostly good to go! I left two questions, and also ran into some wonkiness where one of the donation block buttons clips into the side rail ad.
Screenshot 2025-08-07 at 14 41 14

includes/providers/gam/class-gam-scripts.php
miguelpeixe: It's intentional, notice that this variable is already checked on line 240/

includes/class-core.php
miguelpeixe: That file is automatically generated by [`@wordpress/scripts`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/) when building the asset, so it should always accompany the asset itself. It's meant to provide accurate enqueue dependencies and a convenient hash version.

includes/providers/gam/class-gam-scripts.php
chickenn00dle: Ha. Just noticing. Thanks for clearing that up 🤦

miguelpeixe:
> ran into some wonkiness where one of the donation block buttons clips into the side rail ad.

Which page template is this page using? It shouldn't be colliding with other elements...

chickenn00dle (approved):
> Which page template is this page using? It shouldn't be colliding with other elements...

This was the default "with sidebar" template. But its strange, I just tested again and this time the ad and page cleanly resize... Might have been a one-time issue with my browser. In any case, since I can no longer reproduce, gonna approve.

PR #978, opened 2025-07-30
feat(gam): support MCM (parent network code)

miguelpeixe:
To be reviewed and merged after #967 

---

Implements support for [Google MCM](https://support.google.com/admanager/answer/11130475) via an optional parent network code:

image

The wizard was updated to include a Save button instead of updating on the select changes since it now includes a text field.

If set, the parent network code should be used in the slot configuration like so: `{parentNetworkCode},{networkCode}`

### Testing

1. Confirm you have GAM connected and some placements configured
2. Checkout this branch and confirm your ads continue to render without issues
3. Navigate to Advertising -> Providers and configure GAM
4. Confirm the new text input renders and shown in the image above
5. Set a random number and save
6. Visit the frontend and look for the `ad_config` js variable in the markup
7. Confirm the `network_code` property holds `{parentNetworkCode},{networkCode}`

### DIFF
--- a/includes/providers/gam/class-gam-model.php
+++ b/includes/providers/gam/class-gam-model.php
@@ -31,6 +31,8 @@ final class GAM_Model {
 
 	const OPTION_NAME_DEFAULT_UNITS = '_newspack_ads_gam_default_units';
 
+	const OPTION_NAME_PARENT_AD_UNIT = '_newspack_ads_gam_parent_ad_unit';
+
 	/**
 	 * GAM Api
 	 *
@@ -343,7 +345,8 @@ final class GAM_Model {
 		if ( true === $sync ) {
 			$api = self::get_api();
 			if ( $api ) {
-				$gam_ad_units = $api->ad_units->get_serialized_ad_units( [], true );
+				$parent_ad_unit_id = self::get_parent_ad_unit_id();
+				$gam_ad_units      = $api->ad_units->get_serialized_ad_units( $parent_ad_unit_id, [], true );
 				if ( ! is_wp_error( $gam_ad_units ) ) {
 					foreach ( $ad_units as $ad_unit_key => $ad_unit_config ) {
 						/** Validate config before continuing. */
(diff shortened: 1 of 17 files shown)


adekbadek (approved):

includes/providers/gam/class-gam-model.php
```diff
@@ -438,13 +440,22 @@ private static function get_legacy_ad_units() {
...
 	/**
 	 * Get parent ad unit id.
 	 *
-	 * @return int|false Parent ad unit id or false if not set.
+	 * @return string Parent ad unit id or empty string if not set.
 	 */
 	public static function get_parent_ad_unit_id() {
-		return get_option( self::OPTION_NAME_PARENT_AD_UNIT, false );
+		return get_option( self::OPTION_NAME_PARENT_AD_UNIT, '' );
 	}
```
rbcorrales, line 459: The change here couldn't potentially cause issues in `add_ad_unit()`?
```php
$parent_ad_unit_id = self::get_parent_ad_unit_id();
if ( $parent_ad_unit_id ) {
```
It might be better to change it to `if ( ! empty( $parent_ad_unit_id ) )`.

Are there any other cases where this could break backward compatibility?

rbcorrales (changes requested):

includes/providers/gam/class-gam-model.php
miguelpeixe: What's the issue in `add_ad_unit()`?

rbcorrales: Nvm, followed all the function usages and it doesn't seem it would cause issues.

rbcorrales (approved):

includes/providers/gam/class-gam-model.php
rbcorrales: And sorry for the bad example! 😄

miguelpeixe: Did you happen to look at https://github.com/Automattic/newspack-plugin/pull/4117 too?

miguelpeixe: No worries! Thank you for being thorough!

PR #984, opened 2025-08-12
fix(placements): side rail placement dimension race condition

miguelpeixe:
If the ad renders before the side rail placement script, it won't catch the rendered ad dimensions and fail to handle collisions properly.

This PR tweaks the reference node to be more restrictive and also initializes it with the ad dimensions, in case it has already rendered.

### Testing

1. In release, put [the script initialization](https://github.com/Automattic/newspack-ads/blob/5094654a6625dc089661e092ec74ec782128d5f7/src/frontend/side-rail-placements.js#L131-L136) in a long  setTimeout() - enough to run after the ad renders
2. Visit a page that renders the side rail placements, adjust your browser height to be smaller than the rendered creative
3. Confirm the ad never gets hidden and is partially visible
4. Checkout this branch, apply the same setTimeout() and confirm that collision is properly handled once the script runs

### DIFF
--- a/includes/integrations/class-side-rail-placements.php
+++ b/includes/integrations/class-side-rail-placements.php
@@ -24,6 +24,7 @@ class Side_Rail_Placements {
 		add_action( 'init', [ __CLASS__, 'register_placements' ] );
 		add_filter( 'newspack_ads_gtag_ads_data', [ __CLASS__, 'filter_ad_units' ] );
 		add_filter( 'newspack_ads_placement_classnames', [ __CLASS__, 'filter_classnames' ], 10, 2 );
+		add_filter( 'newspack_ads_gam_ad_unit_initial_display', [ __CLASS__, 'filter_ad_unit_initial_display' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


chickenn00dle (commented):
I'm seeing some strange flicker behavior now. Looks like tall ads load initially in a collision state, then immediately hide until I scroll.

https://github.com/user-attachments/assets/0c3e6b78-9fa2-4c76-8e3f-6518f4cadf91

This is on Chrome.

miguelpeixe:
Thanks for catching that!

I've introduced an initial display management to avoid the flickering; it should be good now.

miguelpeixe:
To support a larger height, I also pushed 489d345afd3b5949c846dbb7fe198c38331c61fc to improve the handling of the sticky footer.

The placement no longer has a fixed bottom of 150px and adjusts according to the sticky footer placement. Once the sticky closes, the spacing is adjusted so the ad can render if it fits.

chickenn00dle (approved):
Thanks for the changes! This looks good now.

miguelpeixe:
Thanks for the reviews!

PR #989, opened 2025-08-28
Alpha release Aug 28

miguelpeixe:
A new alpha release.


miguelpeixe:
Skipping this one as #974 was released as a hotfix

PR #1023, opened 2026-02-20
chore(gam): update api version

miguelpeixe:
Update the GAM API version from v202505 to v202511.

## Testing

1. Connect to a GAM account
2. Create, update, and archive an ad unit
3. Confirm all actions work without issues

### DIFF
--- a/includes/providers/gam/api/class-ad-units.php
+++ b/includes/providers/gam/api/class-ad-units.php
@@ -8,19 +8,19 @@
 namespace Newspack_Ads\Providers\GAM\Api;
 
 use Newspack_Ads\Providers\GAM\Api\Api_Object;
-use Google\AdsApi\AdManager\Util\v202505\StatementBuilder;
-use Google\AdsApi\AdManager\v202505\ServiceFactory;
-use Google\AdsApi\AdManager\v202505\InventoryService;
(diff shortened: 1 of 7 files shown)


dkoo (commented):
Tested well with no issues, but the [failing job](https://github.com/Automattic/newspack-ads/actions/runs/22232351905/job/64315132215?pr=1023) is concerning. I'm not able to replicate locally so far.

miguelpeixe:
It's from `php-jwt`, a `google-ads-phplib` dependency. GitHub is preventing the package download via their Security Advisory: https://github.com/firebase/php-jwt/issues/620

The CVE is disputed:

> The "vulnerability" is that php-jwt accepts short HMAC keys without validation.
> This is not a library bug — key management is the caller's responsibility.
> PHP's own hash_hmac() and openssl_sign() behave identically and have no CVEs
> for this.
> NVD agrees — hence the Disputed tag and no score from NIST.

dkoo:
Does that mean we should hold this until the CVE report is resolved?

miguelpeixe:
This is in trunk as well, unrelated to this change.

dkoo (approved):
Let's merge it and let it sort itself out. Hopefully it won't block release workflows in the meantime

miguelpeixe:
Thank you!

PR #1056, opened 2026-04-14
feat: add a warning message when hidden block contains an ad

laurelfulford:
In WordPress 7.0, there's a new option to hide blocks at certain breakpoints.

Since hiding ads with CSS can be problematic and cause issues with ad providers, we've already removed this option from the ad blocks themselves. With this PR:

* If you hide a block at certain breakpoints and it contains an ad block, a warning will display at the top of the editor
* If you hide a block outright that contains an ad block, no warning is displayed - that's because these blocks aren't output at all on the front-end
* On the front-end, if you've hidden a block at a certain breakpoint and it contains an ad block, we use CSS to force it to be visible

This approach is a little subpar - I had wanted to also force the blocks to be visible in the editor, but it looks like they're removed from the output in the editor unless they're focused on. So I hope the notice + force-showing on the front-end is enough! 😬 

## Steps to test

1. Start on a test site running the latest WordPress 7.0 RC.
2. Add a columns or group block to the editor, and then add an ad block inside of it.
3. Click the parent block, select the kebab (`...`) menu in the block toolbar, and select Hide.
4. Check 'Omit from published content'. Confirm that you get no errors/weirdness in the editor, and that the block doesn't appear at all in the markup rendered on the front-end:

CleanShot 2026-04-14 at 10 59 14

5. Repeat step 3, but this time, pick one of the other options (you can check one breakpoint, or all three). Confirm you get an error at the top of the editor:

CleanShot 2026-04-14 at 11 20 15

6. Publish the post/page and confirm this second hidden block is visible at all breakpoints. 
7. Try removing the ad block and confirming the warning goes away in the editor, and that the block's hidden as expected based on your settings on the front-end.
8. Repeat the testing steps on a site running the latest version of WP 6.9. The hide option is just a simple toggle within the dropdown menu (no modal, and no options to choose a breakpoint). It should work just like the "Omit from published content" in 7.0, and do nothing.

### DIFF
--- a/src/frontend/style.scss
+++ b/src/frontend/style.scss
@@ -147,6 +147,6 @@
 .wp-block-hidden-tablet,
 .wp-block-hidden-mobile {
 	&:has(.wp-block-newspack-blocks-wp-block-newspack-ads-blocks-ad-unit) {
-		display: block !important;
+		display: revert !important;
 	}
 }
--- a/src/frontend/style.scss
+++ b/src/frontend/style.scss
@@ -141,3 +141,12 @@
 		}
 	}
 }
+
+// Don't allow hiding blocks containing ad blocks.
+.wp-block-hidden-desktop,
+.wp-block-hidden-tablet,
+.wp-block-hidden-mobile {
+	&:has(.wp-block-newspack-blocks-wp-block-newspack-ads-blocks-ad-unit) {
+		display: block !important;
+	}
+}
--- a/src/frontend/style.scss
+++ b/src/frontend/style.scss
@@ -142,10 +142,11 @@
 	}
 }
 
-// Don't allow hiding blocks containing ad blocks.
+// Force visibility on parent blocks the user has hidden when they contain an ad unit.
 .wp-block-hidden-desktop,
 .wp-block-hidden-tablet,
 .wp-block-hidden-mobile {
+	// This CSS class seems wrong, but please don't fix! The double prefix matches what's output by the plugin.
 	&:has(.wp-block-newspack-blocks-wp-block-newspack-ads-blocks-ad-unit) {
 		display: revert !important;
 	}
--- a/src/setup/visibility-warning.js
+++ b/src/setup/visibility-warning.js
@@ -33,7 +33,7 @@ const AdVisibilityWarning = () => {
 					'A hidden block contains an ad unit. Ads and their containers must stay visible at all breakpoints to avoid penalties — use your ad provider settings to control ad visibility. Find hidden blocks in Document Overview > List view.',
 					'newspack-ads'
 				),
-				{ id: NOTICE_ID, isDismissible: true }
+				{ id: NOTICE_ID, isDismissible: false }
 			);
 		} else {
 			removeNotice( NOTICE_ID );
--- a/src/setup/visibility-warning.js
+++ b/src/setup/visibility-warning.js
@@ -20,7 +20,7 @@ const AdVisibilityWarning = () => {
 			if ( getBlockName( id ) !== AD_BLOCK ) {
 				return false;
 			}
-			return getBlockParents( id ).some( isHidden );
+			return [ id, ...getBlockParents( id ) ].some( isHidden );
 		} );
 	}, [] );
 
--- a/src/setup/visibility-warning.js
+++ b/src/setup/visibility-warning.js
@@ -30,18 +30,11 @@ const AdVisibilityWarning = () => {
 	useEffect( () => {
 		if ( hasHiddenAdContainer ) {
 			createWarningNotice(
-				'<p>' +
-					__(
-						'One or more hidden blocks contain an ad unit. Ad blocks and their containers will remain visible on all screen sizes — hiding ads with CSS may result in penalties.',
-						'newspack-ads'
-					) +
-					'</p><p>' +
-					__(
-						'To control which ads appear at different breakpoints, use your ad provider settings. To find the hidden block, go to Document Overview > List view.',
-						'newspack-ads'
-					) +
-					'</p>',
-				{ id: NOTICE_ID, isDismissible: true, __unstableHTML: true }
+				__(
+					'One or more hidden blocks contain an ad unit. Ad blocks and their containers will remain visible on all screen sizes — hiding ads with CSS may result in penalties. To control which ads appear at different breakpoints, use your ad provider settings. To find the hidden block, go to Document Overview > List view.',
+					'newspack-ads'
+				),
+				{ id: NOTICE_ID, isDismissible: true }
 			);
 		} else {
 			removeNotice( NOTICE_ID );
--- a/src/setup/visibility-warning.js
+++ b/src/setup/visibility-warning.js
@@ -31,7 +31,7 @@ const AdVisibilityWarning = () => {
 		if ( hasHiddenAdContainer ) {
 			createWarningNotice(
 				__(
-					'One or more hidden blocks contain an ad unit. Ad blocks and their containers will remain visible on all screen sizes — hiding ads with CSS may result in penalties. To control which ads appear at different breakpoints, use your ad provider settings. To find the hidden block, go to Document Overview > List view.',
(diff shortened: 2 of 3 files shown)


src/frontend/style.scss
```diff
@@ -141,3 +141,12 @@
 		}
 	}
 }
+
+// Don't allow hiding blocks containing ad blocks.
+.wp-block-hidden-desktop,
+.wp-block-hidden-tablet,
+.wp-block-hidden-mobile {
```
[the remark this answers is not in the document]
laurelfulford: `:has()` has great coverage in modern browsers - this is fine. https://caniuse.com/?search=%3Ahas%28%29

src/setup/visibility-warning.js
```diff
@@ -0,0 +1,54 @@
...
+						'newspack-ads'
+					) +
+					'</p><p>' +
+					__(
+						'To control which ads appear at different breakpoints, use your ad provider settings. To find the hidden block, go to Document Overview > List view.',
+						'newspack-ads'
+					) +
+					'</p>',
+				{ id: NOTICE_ID, isDismissible: true, __unstableHTML: true }
+			);
```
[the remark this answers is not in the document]
laurelfulford: Updated in [738941a](https://github.com/Automattic/newspack-ads/pull/1056/commits/738941a95219e1f49917521dc58451a0e0509f6e) and [c6b80f2](https://github.com/Automattic/newspack-ads/pull/1056/commits/c6b80f289002f544b404d6e8d3af8e42557ee263).

src/setup/visibility-warning.js
```diff
@@ -0,0 +1,46 @@
...
+	const { createWarningNotice, removeNotice } = useDispatch( 'core/notices' );
+
+	useEffect( () => {
+		if ( hasHiddenAdContainer ) {
+			createWarningNotice(
+				__(
+					'A hidden block contains an ad unit. Ads and their containers must stay visible at all breakpoints to avoid penalties — use your ad provider settings to control ad visibility. Find hidden blocks in Document Overview > List view.',
+					'newspack-ads'
+				),
+				{ id: NOTICE_ID, isDismissible: true }
```
adekbadek: Worth considering `isDismissible: false`. With a stable id and `isDismissible: true`, once the user clicks the X, the notice store drops the entry, and because the `useEffect` only re-runs when `hasHiddenAdContainer` flips, the user can then toggle which breakpoints are hidden, add more ad blocks inside hidden parents, save, refresh — and never see the warning again. I confirmed this empirically. The only way to re-trigger is to fully un-hide and then re-hide.

The front-end CSS override still kicks in regardless, so this isn't fully load-bearing — just thinking the editor warning loses much of its value once dismissed.

src/setup/visibility-warning.js
```diff
@@ -0,0 +1,46 @@
...
+		} );
+	}, [] );
+
+	const { createWarningNotice, removeNotice } = useDispatch( 'core/notices' );
+
+	useEffect( () => {
+		if ( hasHiddenAdContainer ) {
+			createWarningNotice(
+				__(
+					'A hidden block contains an ad unit. Ads and their containers must stay visible at all breakpoints to avoid penalties — use your ad provider settings to control ad visibility. Find hidden blocks in Document Overview > List view.',
```
adekbadek: Heads-up on i18n: `Document Overview > List view` concatenates UI labels that core translates per locale, so non-English users will see two different translations of the menu chrome. Options: drop the locator hint, or pull the labels from core (`__( 'Document Overview', 'default' )` etc.). Confirmed the English string matches WP 7.0's actual UI (modulo `View` capitalization).

src/setup/visibility-warning.js
```diff
@@ -0,0 +1,46 @@
...
+const AdVisibilityWarning = () => {
+	const hasHiddenAdContainer = useSelect( select => {
+		const { getClientIdsWithDescendants, getBlockName, getBlockParents, getBlockAttributes } = select( 'core/block-editor' );
+		const isHidden = clientId => {
+			const viewport = getBlockAttributes( clientId )?.metadata?.blockVisibility?.viewport;
+			return viewport && Object.values( viewport ).some( visible => visible === false );
+		};
+		return getClientIdsWithDescendants().some( id => {
+			if ( getBlockName( id ) !== AD_BLOCK ) {
+				return false;
```
adekbadek, line 21: Nit: `getBlockParents` doesn't include the block itself, so if `metadata.blockVisibility` is ever applied directly to an ad-unit (e.g. via patterns or a future UI affordance), this won't fire. Cheap to harden: `[ id, ...getBlockParents( id ) ].some( isHidden )`.

src/frontend/style.scss
```diff
@@ -141,3 +141,12 @@
 		}
 	}
 }
```
adekbadek, line 143: Nit: this comment reads like the rule prevents hiding, but the rule actually overrides the user's hide. Maybe: `// Force visibility on parent blocks the user has hidden when they contain an ad unit.`

src/frontend/style.scss
```diff
@@ -141,3 +141,12 @@
...
+
+// Don't allow hiding blocks containing ad blocks.
+.wp-block-hidden-desktop,
+.wp-block-hidden-tablet,
```
adekbadek, line 147: Nit: this doubled-prefix class looks like a typo but is intentional — `Ad_Unit_Block::block_classes()` prepends `wp-block-newspack-blocks-` to a `$type` that's already `wp-block-newspack-ads-blocks-ad-unit`. Worth a one-line code comment so a future maintainer doesn't "fix" it.

adekbadek (approved):
Tested in an isolated env on WP 7.0-RC1 with the worktree's built newspack-ads. All three bullets from the description verified end-to-end:

- Setting `metadata.blockVisibility.viewport.<bp> = false` on a Columns block containing an ad-unit triggers the warning notice.
- Setting `metadata.blockVisibility = false` (outright hide) does NOT trigger the warning, consistent with the schema split in core's block-editor.
- The compiled CSS rule loads on the front-end. With a synthetically injected ad-unit child in a hidden parent, computed `display` flips from `none` → `block`. The doubled-prefix class is real (verified `Ad_Unit_Block::block_classes()` prepends `wp-block-newspack-blocks-` to the type).

Approving — see inline notes for one higher-impact suggestion (dismissibility) and a few smaller ones. Nothing blocking.

src/setup/visibility-warning.js
laurelfulford: Thanks @adekbadek! I flip flopped about making it dismissable, but the fact that users can get into situations where they can stop seeing the warning is a really good reason not to. Changed in [c83b23c](https://github.com/Automattic/newspack-ads/pull/1056/commits/c83b23ced03a9b90dcf75b621e8d96e7e9a99484)!

src/setup/visibility-warning.js
laurelfulford: Oooo yeah, good call! I went with just removing the locator hint: [cbed902](https://github.com/Automattic/newspack-ads/pull/1056/commits/cbed9023eaa0f3f6ff7308b060cef21000e8003f)

src/frontend/style.scss
laurelfulford: Updated in [27ede98](https://github.com/Automattic/newspack-ads/pull/1056/commits/27ede98bd82612ee20205a188128a31d240ca2ab)!

src/frontend/style.scss
laurelfulford: That CSS class always makes me laugh, but this is a really good point - I added a comment to flag it as "correct" just in case someone thinks its a typo in [27ede98](https://github.com/Automattic/newspack-ads/pull/1056/commits/27ede98bd82612ee20205a188128a31d240ca2ab).

src/setup/visibility-warning.js
laurelfulford: Good call! Added in [7ddafc9](https://github.com/Automattic/newspack-ads/pull/1056/commits/7ddafc9939c515f33a7dfd44a4ae6161a8a7dff0).

laurelfulford:
Thanks for the review and helpful suggestions, @adekbadek! 🙌

PR #1061, opened 2026-04-28
fix(tabs): align tab-header overlay precisely over its button

miguelpeixe:
Follow-up to #1031, fixing the tab-header rendering glitch, present in `trunk` as well. When clicking back on an existing tab, the editable header overlay would appear duplicated, misaligned with the tab button, and only "snap" into place once typing forced a re-render.

The strategy has been refactored to track each button's bounding rect with a ResizeObserver and write the position synchronously via useLayoutEffect so the overlay stays in register through wraps, viewport resizes, and tab visibility changes. It also centers the overlay's rich-text with flex to match the button's flex centering, eliminating the vertical drift on multi-line tabs.

## Testing

1. Draft a post and add the Tabs block
2. Resize your browser window so it's narrower
3. Add multiple tabs, make sure some of them wrap
4. Switch between them, editing the tab header text
5. Make sure the editable text is always snapped in the correct position

### DIFF
--- a/src/blocks/tabs/edit.js
+++ b/src/blocks/tabs/edit.js
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
  */
 import { createBlock } from '@wordpress/blocks';
 import { compose, ifCondition, useRefEffect } from '@wordpress/compose';
-import { useState, useEffect, useCallback, Fragment } from '@wordpress/element';
+import { useState, useEffect, useLayoutEffect, useCallback, Fragment } from '@wordpress/element';
 import { withSelect, withDispatch } from '@wordpress/data';
 import { Button, NavigableMenu } from '@wordpress/components';
 import { plus } from '@wordpress/icons';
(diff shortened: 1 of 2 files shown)


laurelfulford (changes requested):
Thanks @miguelpeixe! This looks good overall. 

The only case I can still get the fuzzy text in the editor is if I get a tab wrapping to more than one line and type a space or a few spaces - the spaces show in the RichText editor but isn't carried over to the Button's actual text:

Image

It looks like if you edit the editor.scss and add the style `white-space: pre-wrap;` to the selector `& .components-button.newspack-ads__tab-item` (at line 101) that should fix it!

Otherwise the issues I spotted seem to be preexisting (like the tabs don't wrap on the front-end, but they're like that on trunk, too). This block might be due a rework just to get the front-end and editor more synced up, but that's well beyond the purpose of this PR! Once the space issue is fixed this should be good to go!

miguelpeixe:
Good catch, @laurelfulford!

Thank you for pointing in the right direction. The suggested fix did the trick. Added in d6e0a885b390f57f4294cf3db5ee2f06bbcdfe4b

laurelfulford (approved):
Looks good, @miguelpeixe - thank you! 🚢

